PR #82, opened 2018-02-27
Preferences option linux

inderpreetsingh:
Shows preferences option in Linux. Goes hand in hand with <pr link here>


inderpreetsingh:
Rebased from dev and sent another PR.

Issue #69, opened 2018-02-03
Preference Option is not showing in the linux


sukhpalsingh:
The preference option in the file menu does not show up in linux. Not sure if this is an issue in window or mac as well. Others options do show up.

tarunsingh5:
Confirmed this behavior is happening

tarunsingh5:
@inderpreetsingh this was addressed in your pull request right?

inderpreetsingh:
Yes, at-least partially it's solved. Preferences menu shows up in linux but it does not show up from `file->preferences`.

tarunsingh5:
@inderpreetsingh can we fix this fully?

inderpreetsingh:
@tarunsingh5 ha'nji it's already fixed. We can close this issue.

PR #83, opened 2018-02-27
Shows preferences option in linux

inderpreetsingh:
Shows preferences option in linux, goes hand in hand with https://github.com/KhalisFoundation/sttm-core/pull/7


inderpreetsingh:
This solves the issue #69

Issue #38, opened 2018-01-12
Help section


tarunsingh5:
Provide ability to open help guide, tutorials, etc.

PR #86, opened 2018-03-01
Adding help section

inderpreetsingh:
In order to solve issue #38  I ended up doing couple of stuff that I believe will help us in adding future markdown based  static content windows.

* I refactored `openChangeLogWindow` function into `openSecondaryWindow` so that it can open any window we want it to by just passing an argument.
* Then used `openSecondaryWindow` to open a help/guide window.
* Refactored `changelog.js`  into `markdownToHTML` so that it can convert any markdown file into html and load into respective id.
* Refactored `changlog.scss` into `markdown.scss` to allow it to style all the markdown based files
* Added content into newly created `HELP.md`
* Added some typographic rules and vertical rhythm into `markdown.scss`. Now both changelog window and help window look beautiful <3 <3 
For now I am loading images and videos from website to not bloat the app size, however let me know if we want to load them into app.

### DIFF
--- a/HELP.md
+++ b/HELP.md
@@ -1,0 +1,53 @@
+###  Video Tutorial
+
+<div class="videoWrapper"><iframe width="100%" height="400px" src="https://www.youtube.com/embed/ZDX8nPkDBSc" frameborder="0" allow="autoplay; encrypted-media" allowfullscreen=""></iframe></div>
+
+### How do I install the desktop application?
+
+You can install SikhiToTheMax (STTM) by visiting [this website](https://khalisfoundation.org/portfolio/sikhitothemax-everywhere/) and choosing the Windows or macOS download link. After that, open the installer and follow the steps that are displayed. ![](https://www.sikhitothemax.org/assets/images/help/desktop-search-ex.png)
+
+### How do I search for a shabad?
+
+After launching STTM, by default you can search for a shabad by entering the first letter of each word. For example, if the shabad is ਗੁਰੁ ਮੇਰੈ ਸੰਗਿ ਸਦਾ ਹੈ ਨਾਲੇ , you would enter "gmsshn".'" ![](https://www.sikhitothemax.org/assets/images/help/desktop-search-ex.png) Alternatively, you can click on the gurmukhi keyboard icon and type in the letters manually. ![](https://www.sikhitothemax.org/assets/images/help/desktop-keyboard.png)
+
+### What do the different search types do?
+
+*   First Letter (Start)
+    *   The letters you input must be from the START of a line in a shabad.
+*   First Letter (Anywhere)
+    *   The letters you input can be from any part of the line. For example, you can start your search by typing in the first letter of every word in the second half of a shabad’s line.
+*   Full Word (Gurmukhi)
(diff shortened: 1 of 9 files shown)


HELP.md
```diff
@@ -0,0 +1,53 @@
+###  Video Tutorial
+
+<div class="videoWrapper"><iframe width="100%" height="400px" src="https://www.youtube.com/embed/ZDX8nPkDBSc" frameborder="0" allow="autoplay; encrypted-media" allowfullscreen=""></iframe></div>
```
navdeepsinghkhalsa: Could you please use "lowercase-with-dashes" for CSS classes and id's?

www/js/controller.js
```diff
@@ -168,10 +168,16 @@ const winMenu = [
         enabled: false,
       },
       ...updateMenu,
+      {
+        label: 'Guide...',
```
navdeepsinghkhalsa: Could we please put this in a Help menu?

www/js/controller.js
```diff
@@ -188,10 +194,16 @@ const macMenu = [
         role: 'about',
       },
       ...updateMenu,
+      {
+        label: 'Guide...',
```
navdeepsinghkhalsa: Same as above - under a new Help top-level menu

www/src/scss/_markdown.scss
```diff
@@ -0,0 +1,65 @@
+$container: 700px;
+$base-font-size: 16px;
+$ratio:1.5;
```
navdeepsinghkhalsa: Could you please add a space after the colon?

www/src/scss/_markdown.scss
```diff
@@ -0,0 +1,65 @@
...
+  margin: auto;
+  max-width: $container;
+  padding: $vertical-rhythm;
+
+  h2,
+  h3,
+  h4 {
+    color: #333;
+    line-height: 1;
+    margin-bottom: $vertical-rhythm/2;
```
navdeepsinghkhalsa: Could you please use spaces between operators?

navdeepsinghkhalsa: All other instances as well

www/help.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <title>Changelog</title>
```
navdeepsinghkhalsa: Could you please change the title? It should be `SikhiToTheMax Help`

www/js/controller.js
inderpreetsingh: I think this is already done!

navdeepsinghkhalsa: Yes, you're right. Sorry

www/js/controller.js
navdeepsinghkhalsa: Could this please be moved to match macOS menubar? With a top-level menu called Help

navdeepsinghkhalsa:
@inderpreetsingh Just one change left for this one and we're good to merge!

inderpreetsingh:
@navdeepsinghkhalsa veerji, I don't have mac so wasn't able to test. I hope it works fine!

PR #95, opened 2018-03-09
changes font color of text in dark theme

inderpreetsingh:
changes font color of translation in dark theme.

For future reference: We need to restructure the scss and theme files to prevent such things from happening.


navdeepsinghkhalsa:
@inderpreetsingh Travis indicates that this PR has a linting error. Could you please fix and update?

PR #102, opened 2018-03-29
Custom theme

inderpreetsingh:
* Created a new type of page called options-page
* We don't want to clutter the navigator area anymore to avoid confusions (might even move some of the existing things out of it), so this is a new interface area that slides from right and shows options about some setting etc.
* Added options page for custom theme and separate localstorage for it.
* Cleans up menu.js a little bit.
* Goes hand in hand with https://github.com/KhalisFoundation/sttm-core/pull/13
customtheme

### DIFF
--- a/app.js
+++ b/app.js
@@ -280,6 +280,12 @@ ipcMain.on('update-settings', () => {
   }
 });
 
+ipcMain.on('update-theme', () => {
+  if (viewerWindow) {
+    viewerWindow.webContents.send('update-theme');
+  }
+});
+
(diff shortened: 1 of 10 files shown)


tarunsingh5:
@navdeepsinghkhalsa can you please review?

navdeepsinghkhalsa:
There is a UI discussion that started on Slack

tarunsingh5:
I think this is a step up from what we have and is something users expect -
I've been asked countless times in person about hey why can't I change the
background color to blue etc

PR #111, opened 2018-04-04
adds ang search and romanized gurmukhi search

inderpreetsingh:
* Goes hand in hand with https://github.com/KhalisFoundation/sttm-core/pull/14
angsearch
romanizedgurmukhi

## Suggestions
* Phonetic search in romanized gurmukhi search.
* Separate select box for selecting source in search.

### DIFF
--- a/www/js/search.js
+++ b/www/js/search.js
@@ -38,21 +38,21 @@ module.exports = {
         break;
       }
       case 2: // Full word (Gurmukhi)
-      case 3: { // Full word (English)
+      case 3: // Full word (English)
+      case 4: { // Full word (Transliteration)
         if (searchType === 2) {
           searchCol = 'v.Gurmukhi';
-        } else {
+        } else if (searchType === 3) {
           searchCol = 'v.English';
+        } else {
+          searchCol = 'v.Transliteration';
         }
-        const words = searchQuery.split(' ');
-        dbQuery = `%${words.join(' %')}%`;
-        condition = `${searchCol} LIKE '${dbQuery}'`;
         break;
       }
       case 5: // Ang
         searchCol = 'PageNo';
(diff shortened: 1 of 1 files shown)


www/js/search.js
```diff
@@ -38,21 +38,21 @@ module.exports = {
...
-        const words = searchQuery.split(' ');
-        dbQuery = `%${words.join(' %')}%`;
-        condition = `${searchCol} LIKE '${dbQuery}'`;
         break;
       }
       case 5: // Ang
         searchCol = 'PageNo';
         dbQuery = parseInt(searchQuery, 10);
-        condition = `${searchCol} = ${dbQuery} AND v.SourceID = '${global.platform.search.currentMeta.source}'`;
+        condition = `${searchCol} = ${dbQuery} AND v.SourceID = '${global.core.search.currentMeta.source || 'G'}'`;
```
inderpreetsingh: @navdeepsinghkhalsa I was not sure why `platform.search.currentMeta` was here. I could find `currentMeta` in core.search but it didn't have a source attribute. So I used 'G' as default source.

navdeepsinghkhalsa:
@inderpreetsingh This is a good PR, but our technique for transliteration search is a bit different than just comparing to the transliteration column, because everybody has different spellings. We'll update this after merged codebase

inderpreetsingh:
@navdeepsinghkhalsa Veerji, ha'nji was thinking the same. What do you think about using some phonetic library? I have had some success previously with phonetic search, when searching for romanized gurmukhi. 🙏 

Also, if you can point me to what technique we are using, I can try using that one in this PR.

inderpreetsingh:
Closing this PR as ang search has been sent in another PR and romanized search is yet to be finalized.

PR #117, opened 2018-04-17
Custom theme and backgrounds 🎨 🖌 

inderpreetsingh:
🎨
* replaces earlier theme editor with pre-built themes from aman veerji  
* you can still add your own themes but by going into themes.json

🖼
* adds custom backgrounds option
* all the backgrounds added are in "Do what you want" License. 
* users can add their backgrounds by adding images to assets/custom_backgrounds folder

🎉 
* adds color overlay to ensure readability of text 
* cleans up menu.js a bit

I know adding images increases the app size but in contrast to loading images from internet every time, I thought this was a better option.


www/js/menu.js
```diff
@@ -2,57 +2,68 @@ const h = require('hyperscript');
 const settings = require('./settings');
 const getJSON = require('get-json');
 
+const buttonFactory = (options) => {
+  let classList;
+  const btnId = options.buttonId || '';
+  const btnIcon = options.buttonIcon || 'fa-times';
```
bogas04: You can destructure options and set defaults there

```diff
- const buttonFactory = (options) => {
+ const buttonFactory = ({ buttonId = '', buttonIcon = 'fa-times' }) => {
  let classList;
-  const btnId = options.buttonId || '';
-  const btnIcon = options.buttonIcon || 'fa-times';
...
```

www/js/menu.js
```diff
@@ -2,57 +2,68 @@ const h = require('hyperscript');
...
     {
       onclick: () => {
         global.platform.search.randomShabad()
-          .then((shabadId) => {
-            global.core.search.loadShabad(shabadId);
-            module.exports.toggleMenu('#shabad-menu-page');
-            // go to shabad page
-            document.querySelector('#shabad-pageLink').click();
-          });
+          .then((shabadId) => { goToShabadPage(shabadId); });
```
bogas04: No need to create extra function :D 

```diff
- .then((shabadId) => { goToShabadPage(shabadId); });
+ .then(goToShabadPage);
```

www/js/viewer.js
```diff
@@ -91,6 +97,11 @@ global.platform.ipc.on('update-settings', () => {
   core.menu.settings.applySettings(prefs);
 });
 
+global.platform.ipc.on('update-theme', () => {
+  const userTheme = JSON.parse(window.localStorage.getItem('customTheme'));
```
bogas04: Make sure you add try/catch block around JSON.parse/stringify just to be safe.

www/js/theme_editor.js
```diff
@@ -0,0 +1,89 @@
...
+
+const imagesPath = 'assets/custom_backgrounds';
+
+const defaultTheme = customThemes[0];
+defaultTheme.bgImage = '';
+
+const getCurrentTheme = () => {
+  const currentThemeString = localStorage.getItem('customTheme');
+  if (currentThemeString) {
+    return JSON.parse(currentThemeString);
```
bogas04: Make sure you try/catch around JSON.parse

www/js/menu.js
bogas04: Sorry, I didn't get your point. Basically right now a new function is created which takes the argument and gives it to goToShabadPage. For such scenario, we can simply pass function reference.

inderpreetsingh: Oh! Okay, got it! Thank you 🙏

PR #121, opened 2018-05-01
27 shabad auto navigation

sukhpalsingh:
Changing key shorts for up and down keys as follows:

1. Pressing Up will take to the home line.
2. Pressing Down will take to the first line and then each Down key press will take to the next line.


ManjotS:
I don’t agree with changing up and down. Unfortunately keyboards have been
handicapped and we can’t utilize page up and down (thanks apple).

Perhaps we can use h for home and s to go back to shabad line. right and
left are also options for this use case.

sukhpalsingh:
fair enough. can we use 's' and 'f' keys then this is the combination used as controllers for many console games also as forward and backward.

navdeepsinghkhalsa:
@inderpreetsingh 'j' and 'k' are more common in productivity software, but I don't quite find this usage case intuitive. I would like to hear from @saintsoldierx

navdeepsinghkhalsa:
@sukhpalsingh This doesn't seem to work for me

PR #147, opened 2018-05-28
Store theme under app.theme property 

tsingh777:
For casting themes

image


www/js/desktop_scripts.js
```diff
@@ -114,7 +114,10 @@ module.exports = {
 
   updateTheme() {
     global.webview.send('update-theme');
-    global.platform.ipc.send('update-theme');
```
navdeepsinghkhalsa: Please leave in IPC and also add for removeTheme

PR #153, opened 2018-05-30
Add Notifications bell

tsingh777:
Call `http://api.sikhitothemax.org/messages/desktop/` to load notifications.

Reviewers:
@inderpreetsingh 
@navdeepsinghkhalsa 

notifications
image

### DIFF
--- a/www/js/menu.js
+++ b/www/js/menu.js
@@ -1,6 +1,22 @@
 const h = require('hyperscript');
 const settings = require('./settings');
 const getJSON = require('get-json');
+const tingle = require('./vendor/tingle');
+const request = require('request');
+const moment = require('moment');
+const electron = require('electron');
+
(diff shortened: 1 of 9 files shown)


www/js/menu.js
```diff
@@ -29,6 +45,45 @@ const goToShabadPage = (shabadId) => {
...
+    scripts[i].parentNode.removeChild(scripts[i]);
+    i -= 1;
+  }
+  return div.innerHTML;
+};
+
+const scriptTagCheckRegEx = /<[^>]*script/i;
+
+const parseContent = (contentString) => {
+  if (scriptTagCheckRegEx.test(contentString)) {
```
tsingh777: Stripping script tags might be over kill....if we trust the source then we can leave this check/filter out

PR #168, opened 2018-06-13
Send entire Line to obs file. Fixes for printShabad method

tsingh777:
This alters the technique to update the Obs Html file by sending the entire line object via the socket. 

This fixes it for Random Shabad and Anand Karaj as well. 

overlays


tsingh777:
@navdeepsinghkhalsa It Did, Last commit should covers those.

PR #175, opened 2018-06-15
OBS

navdeepsinghkhalsa:


bogas04:
What is this feature about?

package.json
```diff
@@ -68,7 +74,9 @@
     "request": "^2.81.0",
     "request-progress": "^3.0.0",
     "scroll": "^2.0.1",
-    "sqlite3": "^3.1.13",
+    "socket.io": "^2.1.1",
+    "sqlite3": "^3.1.8",
+    "tippy.js": "^2.5.2",
```
tsingh777: where are we using this?

www/js/overlay.js
```diff
@@ -0,0 +1,172 @@
...
+    bottomLayoutApply();
+    break;
+  case 'split':
+    splitLayoutApply();
+    break;
+  default :
+    break;
+}
+updateContentBarColor(overlayVars.bgColor, 'backgroundColor');
+updateContentBarColor(overlayVars.textColor, 'color');
```
tsingh777: Not related to this line but remove the `overlay.json` file?

www/obs/socket.io.js
```diff
@@ -0,0 +1,8 @@
+/*!
+ * Socket.IO v2.1.1
```
tsingh777: We can probably use this for sync? Nothing to change right now but clean up in the future. cc @bogas04

bogas04: Yeah we can use it, I dynamically load socket.io lib from our server as that takes care of connection and everything. We can definitely use the library and connect manually

PR #187, opened 2018-06-20
Changes title Overlay to Bani Overlay

inderpreetsingh:


www/js/controller.js
```diff
@@ -296,6 +290,12 @@ const macMenu = [
           electron.shell.openExternal('https://khalisfoundation.org/donate/');
         },
       },
+      {
+        label: 'Bani Overlay',
```
navdeepsinghkhalsa: I don't think under Help for the macOS app is an intuitive place to have this. @saintsoldierx could you please advise under which menu item Bani Overlay should be on macOS?

PR #195, opened 2018-07-08
adds notification with file path for live feed

inderpreetsingh:
screen shot 2018-07-08 at 6 16 53 pm

We  already had `noty` library ready to take on notifications. So I used that library to show the information about live feed. It has timeout of 7seconds and can be closed anytime with close button. Currently it shows on top-right, but it can also be shown in bottom-right (close to where we shift live feed option). The reason I kept it top-right is because 
a) that's default for most of apps, so user expects to look at top right for notifications 
b) it doesn't obstruct user's workflow in navigator (for those who already know where the files are).

🙏

### DIFF
--- a/www/js/controller.js
+++ b/www/js/controller.js
@@ -1,5 +1,6 @@
 /* global Mousetrap */
 const electron = require('electron');
+const Noty = require('noty');
 
 const remote = electron.remote;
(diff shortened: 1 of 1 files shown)


inderpreetsingh:
closing this until we change default path.

PR #202, opened 2018-07-14
182 : Present Next Line in different style

lbk003:
- Added opacity for next line. 

- Combined animation for all but the main gurbani line

- 90% font size not done. Take a look at this, if still needed, then please open a separate issue for that, so that this part can be merged.


lbk003:
Fixing the fade-out animation per @saintsoldierx

PR #210, opened 2018-07-21
Add resize for background and text for Overlay

tsingh777:
Should fix text overlapping and give more control to OBS user for sizing overlay. 

overlay

### DIFF
--- a/www/js/defaults.json
+++ b/www/js/defaults.json
@@ -45,7 +45,9 @@
       "overlayVars": {
         "layout": "split",
         "bgColor": "#2c3e50",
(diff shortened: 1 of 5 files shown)


navdeepsinghkhalsa:
Looking good. @tsingh777 Is there a reason we're calculating the height of the bar instead of just adding padding around the content?

tsingh777:
@navdeepsinghkhalsa not really doing much calculating. adjusting the padding would be one way to do it as well. Since we have to rework this feature a bit to allow for more customization, think we can go with this for now.

PR #214, opened 2018-07-21
182 present nex line in a different style

lbk003:
Opacity and animation are working.
90% size isn't. Kindly review and suggest changes
@saintsoldierx @bogas04 @inderpreetsingh


lbk003:
PR should target dev.

PR #242, opened 2018-08-10
[231] Allow Scripture filtering and quick switch between search types

tsingh777:
sf

Changing the behavior to always show the scripture filter. 
For Ang Search, default to G for source ID when ~~not one of ```DASAM_GRANTH, GURDAS_VAARAN, GURU_GRANTH_SAHIB, NAND_LAL_VAARAN:```~~ All Sources
When going back to normal search, use the same scripture source and clear ang search.

### DIFF
--- a/www/js/search-database.js
+++ b/www/js/search-database.js
@@ -76,13 +76,10 @@ module.exports = {
         condition = `${searchCol} = ${dbQuery}`;
 
         switch (global.core.search.currentMeta.source) {
-          case CONSTS.SOURCE_TYPES.DASAM_GRANTH:
-          case CONSTS.SOURCE_TYPES.GURDAS_VAARAN:
-          case CONSTS.SOURCE_TYPES.GURU_GRANTH_SAHIB:
-          case CONSTS.SOURCE_TYPES.GURDAS_JI_VAARAN:
-            angSearchSourceId = global.core.search.currentMeta.source;
+          case null:
             break;
           default:
+            angSearchSourceId = global.core.search.currentMeta.source;
             break;
         }
         condition = `${searchCol} = ${dbQuery} AND v.SourceID = '${angSearchSourceId}'`;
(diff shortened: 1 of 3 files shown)


inderpreetsingh:
I think all we want is to hide scripture in case when source selected is "all" and there's still no search results.

screen shot 2018-08-12 at 2 13 49 pm

Also, now that I think about it, If ang search is going to work with all scriptures, may be when we start ang searching, the scripture dropdown should automatically select Sri Guru Granth Sahib Ji (instead of still showing all). What do you think?

tsingh777:
> I think all we want is to hide scripture in case when source selected is "all" and there's still no search results.

I can fix that. 

>  may be when we start ang searching, the scripture dropdown should automatically select Sri Guru Granth Sahib Ji (instead of still showing all).

So with ang search, we only get results back for DASAM_GRANTH, GURDAS_VAARAN, GURU_GRANTH_SAHIB,  GURDAS_JI_VAARAN.

The others return no results.  
https://www.sikhitothemax.org/ang?ang=1&source=A 
https://www.sikhitothemax.org/ang?ang=1&source=N
https://www.sikhitothemax.org/ang?ang=1&source=R

right now I have it default to GURU_GRANTH_SAHIB when one of those is not selected (need to fix which ones i think). It's not very clear as a user that is what is happening. I don't know if I like switching the filter if all is selected and user starts an ang search, but I agree it needs to be presented better. 

@saintsoldierx  @maneetpaul  @navdeepsinghkhalsa Any suggestion in improving this?

inderpreetsingh:
@tsingh777 may be just don't show them in dropdown when user starts typing ang in ang search?

tsingh777:
So now we hide when results and all sources are selected. When any other source is selected and a combination of filters is selected that gives no results, we keep the filters open. 

This should cover most of the cases and allow for better use of the scripture filter.

inderpreetsingh:
LGTM

PR #243, opened 2018-08-10
Adds display options tab

Gauravjeetsingh:
displaytab

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -110,7 +110,6 @@
               <ul id="list-of-custom-slides"></ul>
             </div>
             <div class="tab-content" id ="display-tab-content">
-              <p>Display Stuff here</p>
             </div>
             <div class="tab-content" id ="other-tab-content">
                 <ul id="list-of-shabad-options"></ul>
--- a/www/index.html
+++ b/www/index.html
@@ -87,9 +87,16 @@
             <span class="nav-header-tab nav-header-text" id="insert-tab">
               <i class="fa fa-desktop"></i> Insert
             </span>
-            <span class="nav-header-tab nav-header-text" id="other-tab">
-                <i class="fa fa-ellipsis-h"></i> Other
-              </span>
+            <div class="list">
+              <select class="more-tabs">
+                <option value="display-tab">
+                  Display
+                </option>
+                <option value="other-tab">
+                  Other
+                </option>
+              </select>
+            </div>
           </div>
           <section class="block-list">
             <div class="tab-content active" id="history-tab-content">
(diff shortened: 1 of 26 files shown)


www/js/desktop_scripts.js
```diff
@@ -148,22 +148,36 @@ Array.from($titleButtons).forEach((el) => {
...
   document.querySelector('.tab-content.active').classList.remove('active');
   tabContent.classList.add('active');
 }
 
-const allTabs = document.getElementsByClassName('nav-header-tab');
-
 Array.prototype.forEach.call(allTabs, (element => {
-  element.addEventListener('click', showTabContent);
+  element.addEventListener('click', () => {
+    const clickedTab = event.currentTarget.id;
```
tsingh777: should `event` not be defined first? Seems to be working but style wise doesn't seem consistent with other event listeners. 

```
event => {
    const clickedTab = event.currentTarget.id;
    showTabContent(clickedTab);
    ...
```

www/src/scss/_header.scss
```diff
@@ -105,3 +118,9 @@
     display: block;
   }
 }
+
+.presenter-view.scale-viewer .more-tabs {
+  @media screen and (max-width: 850px) {
```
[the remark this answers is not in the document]
inderpreetsingh: We can do this specific improvement later, @Gauravjeetsingh can we please complete rest of stuff and get this PR merged?

www/index.html
```diff
@@ -87,9 +87,16 @@
             <span class="nav-header-tab nav-header-text" id="insert-tab">
               <i class="fa fa-desktop"></i> Insert
             </span>
-            <span class="nav-header-tab nav-header-text" id="other-tab">
-                <i class="fa fa-ellipsis-h"></i> Other
-              </span>
+            <div class="list">
```
tsingh777: On Mac it doesn't really look nice.
bug

It also doesn't behave the same as the video in the PR. takes 2 clicks to go to the selection when changing the selection.

@saintsoldierx should comment but my suggestion would be not use a select tag for the extra tags.

inderpreetsingh:
@Gauravjeetsingh can you please take a look at this?

Gauravjeetsingh:
Switched to custom dropdown instead of select

aug 21 2018 11_52 pm

tsingh777:
Agree with @inderpreetsingh 
> Actually, if there's space on the right hand side, I think there should be no select at all. It's only when there's no space on right hand side, and the box is small, only then we should have that dropdown. That will also show selectbox at only edge cases (small screens) and we might tolerate that.

We can follow up in separate PR.

PR #248, opened 2018-08-12
Shows presenter-view without external display

inderpreetsingh:
sttm10

### DIFF
--- a/app.js
+++ b/app.js
@@ -211,6 +211,7 @@ function createViewer(ipcData) {
       });
     });
   }
+  mainWindow.webContents.send('presenter-view');
 }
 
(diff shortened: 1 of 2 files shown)


tsingh777:
Code looks good. This is going to be very useful!

I see one minor issue with casting. When we are in presenter view with single screen the cast selection shows in the viewer window. Ideally it should be in the middle? Not a blocker but if it's an easy fix, I think it would look nicer. Also when we stop the casting we are put into non presenter view, but the presenter viewer toggle is still enabled. Again, not a blocker, but if it can keep the consistency it would be nicer. 

cast

Issue #261, opened 2018-08-14
add padding around announcement text box


ManjotS:
![screen shot 2018-08-14 at 3 41 30 pm](https://user-images.githubusercontent.com/644885/44122425-b1482510-9fd8-11e8-8725-05c73f1c205a.png)

tsingh777:
The button as well. Probably shouldn't be that wide either.

Issue #255, opened 2018-08-14
slider icon doesnt do anything


ManjotS:
![screen shot 2018-08-14 at 2 43 38 pm](https://user-images.githubusercontent.com/644885/44120207-756a7cee-9fd0-11e8-9e60-c01819eae94e.png)

Also, down arrow should be closer to the word scripture to show they are associated.

PR #265, opened 2018-08-15
Fixes anncouncement box style and scripture filter style

tsingh777:
https://github.com/KhalisFoundation/sttm-desktop/issues/261 

image

https://github.com/KhalisFoundation/sttm-desktop/issues/255

image


tsingh777:
For the elements targeted, % isn't necessary. A button can be the same size regardless of the resolution.  min-height/margin also doesn't need a percent.

PR #271, opened 2018-08-17
Show which scripture is selected

tsingh777:
Moving scripture filter to left as well so when selection changes the difference in width does not move the filter around too much. 
scripture


inderpreetsingh:
Perfect veerji, thank you!

Issue #281, opened 2018-08-22
Shabad loads twice if there is a quick double click on the search result


lalaagolaa:
Tested it on 2 devices
![image](https://user-images.githubusercontent.com/20653829/44486881-ce988100-a622-11e8-894f-7eaee1b89935.png)

tarunsingh5:
not fixed in 4.1

ManjotS:
related to  #293

PR #284, opened 2018-08-23
Adds Debounce for shabad selection

tsingh777:
Fixes https://github.com/KhalisFoundation/sttm-desktop/issues/281

uses https://lodash.com/docs/4.17.10#debounce


tsingh777:
sort of. https://css-tricks.com/debouncing-throttling-explained-examples/#article-header-id-11 

500 ms is the time we "debounce" the clicks. so if they keep clicking the shabad, it will only load once in a 500 ms interval. 

If it feels too slow I can decrease the time.

tsingh777:
https://ux.stackexchange.com/questions/95336/how-long-should-the-debounce-timeout-be

> TL;DR
Get user feedback, get user feedback, get user feedback. 500ms is a good starting point.

Issue #251, opened 2018-08-13
text cut off on 4:3


ManjotS:
![screen shot 2018-08-13 at 9 14 05 am](https://user-images.githubusercontent.com/644885/44044056-44a859fc-9ed9-11e8-9870-5ab3002d3648.png)

ManjotS:
![screen shot 2018-08-13 at 9 17 45 am](https://user-images.githubusercontent.com/644885/44044223-c21febd4-9ed9-11e8-91d2-fcf88b7f3fc0.png)

ManjotS:
Seeing this issue  at  gurdwaras with  4:3 projectors (most  of them)

inderpreetsingh:
![image](https://user-images.githubusercontent.com/1131610/47014957-5b664380-d169-11e8-8988-de6a649f75f6.png)

It is not being cut off from side but it gets cut from below.

inderpreetsingh:
However that happens everywhere (in all resolutions) 🐛 

![image](https://user-images.githubusercontent.com/1131610/47015014-8355a700-d169-11e8-98c1-15446729dfb5.png)

PR #285, opened 2018-08-23
Fixes sttm logo on 4:3 screen

Gauravjeetsingh:
This fixes issue #251


www/src/scss/viewer.scss
```diff
@@ -228,7 +228,7 @@ h2 {
 .logo {
   display: none;
   height: 20%;
-  left: 5%;
+  left: 72px;
```
tsingh777: This cuts off the logo for 1920x1200
image

try `95px`? I think it works for most screen resolutions.

ManjotS:
Can we do just do 5-7% from left and bottom for all screens?

www/src/scss/viewer.scss
Gauravjeetsingh: @tsingh777 Veerji! the transform property was causing an issue here, fixed that. Now it will be 12px from the edge on every screen.

PR #303, opened 2018-09-03
allows the use of asterisk wild card in first letter search anywhere

inderpreetsingh:
screen shot 2018-09-03 at 8 41 23 pm

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ SikhiToTheMax Desktop App
 
 Create the app package for your system:
 
- * **macOS** - Run `npm run pack:osx`
+ * **macOS** - Run `npm run pack:mac`
  * **Windows 64-bit** - Run `npm run pack:win`
  * **Windows 32-bit** - Run `npm run pack:win32`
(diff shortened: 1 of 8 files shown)


tarunsingh5:
@tsingh777 is this merged?

inderpreetsingh:
@tarunsingh5  @ManjotS  added support to first letter start as well! So when we are using first letter start, we no longer use the optimised query.  🙏

tarunsingh5:
@inderpreetsingh only when there's a wildcard and first letter start right? otherwise we should use optimized

inderpreetsingh:
@tarunsingh5 yes, if it's asterisk - then we use non optimised otherwise optimised.

tarunsingh5:
@navdeepsinghkhalsa does this need to be closed?

navdeepsinghkhalsa:
@tarunsingh5 @inderpreetsingh Let's put a pin in this. It will have to be written in Realm anyway, which is near merging with dev, so (forgive me) this may be redundant to do in SQLite

navdeepsinghkhalsa:
Closing, as we've moved to Realm

PR #313, opened 2018-09-13
Fix 73 consolidate search

navdeepsinghkhalsa:


www/js/banidb/index.js
```diff
@@ -13,6 +12,7 @@ module.exports = {
     // default source for ang search to GURU_GRANTH_SAHIB
     let angSearchSourceId = CONSTS.SOURCE_TYPES.GURU_GRANTH_SAHIB;
     const order = [];
+    let howManyRows = 20;
```
tsingh777: Any specific reason for 20?

www/js/banidb/index.js
```diff
@@ -55,6 +55,7 @@ module.exports = {
       }
       case CONSTS.SEARCH_TYPES.ANG: // Ang
         searchCol = 'PageNo';
+        howManyRows = 1000;
```
tsingh777: is this the max for any ang?

www/js/banidb/index.js
navdeepsinghkhalsa: Just to keep is snappy

www/js/banidb/index.js
navdeepsinghkhalsa: It's just ridiculously high to make sure all cases are covered.

PR #326, opened 2018-09-26
Fix 291 overlay auto adjust

navdeepsinghkhalsa:


navdeepsinghkhalsa:
Thank you. We decided against drag&drop, citing that advanced users will use the text files to customize their layout. Logo implementation will need some thought, but I imagine it would be similar. We can have preset locations with some buttons to tweak spacing

Issue #184, opened 2018-06-19
where are live feed text files when I turn it on? what is it? help has no info.


ManjotS:
Manjot Singh [8 minutes ago]
hover tooltips are good

Manjot Singh [8 minutes ago]
we should also add it to the help (help menu, and on site)

maneetpaul:
Discussion that was had in Slack regarding how this should look. It was decided to add a toggle in the controller with the directory path underneath it. 

We need a way for the user to see whats the current set directory and change it.

@inderpreetsingh is this something you can implement? or is further input needed regarding UI/UX?

tarunsingh5:
#152

tsingh777:
https://github.com/KhalisFoundation/sttm-desktop/pull/328

Issue #280, opened 2018-08-22
Display Options UI Enhancements


trimankandola:
![image](https://user-images.githubusercontent.com/20653829/44484586-2d0e3100-a61c-11e8-964d-9829f67aa556.png)

saintsoldierx:
Please remove from the top left settings and leave in the bottom right panel.

maneetpaul:
Spoke with @saintsoldierx  - see notes below: 

- Move EVERYTHING that is currently in "Preferences" to the "Display Options" tab in the bottom right quadrant of presenter view 
- When somebody goes to "File > Preferences" or presses the keyboard shortcut "Command +," it should go to the "Display Options" tab in the bottom right, we are eliminating the "Settings" menu completely when in Presenter View
- Additionally, remove the dropdown for "Other" in the bottom right quadrant and make it it's own tab next to "Display Options" 

@inderpreetsingh Let us know if this needs any further clarification

tsingh777:
@maneetpaul @saintsoldierx 
> Additionally, remove the dropdown for "Other" in the bottom right quadrant and make it it's own tab next to "Display"
I believe this was done to because we can't fit all the tabs at the top of the box. Think the change here would be how we show extra tabs.

maneetpaul:
Yes for the non-presenter view we should keep the drop-down. But for presenter view, it should be a stand-alone tab.

maneetpaul:
Closing this. Going to address this in a more solidified plan for UI fixes in a future release. Creating a separate issue to address the "Others" as a standalone tab.

PR #328, opened 2018-09-30
Allow live feed file directory selection

tsingh777:
Fixes https://github.com/KhalisFoundation/sttm-desktop/issues/184
image

### DIFF
--- a/app.js
+++ b/app.js
@@ -293,7 +293,8 @@ ipcMain.on('clear-apv', () => {
 });
 
 function createBroadcastFiles(arg) {
-  const userDataPath = electron.app.getPath('userData');
+  const liveFeedLocation = store.get('userPrefs.app.live-feed-location');
+  const userDataPath = (liveFeedLocation === 'default' || !liveFeedLocation) ? electron.app.getPath('desktop') : liveFeedLocation;
   const gurbaniFile = `${userDataPath}/sttm-Gurbani.txt`;
   const englishFile = `${userDataPath}/sttm-English.txt`;
   try {
--- a/www/js/controller.js
+++ b/www/js/controller.js
@@ -2,6 +2,7 @@
 const electron = require('electron');
 
 const remote = electron.remote;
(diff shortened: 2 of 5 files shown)


www/js/controller.js
```diff
@@ -493,4 +494,20 @@ module.exports = {
...
+  livefeed(val) {
+    if (val) {
+      dialog.showOpenDialog({
+        defaultPath: remote.app.getPath('desktop'),
+        properties: ['openDirectory'],
+      }, (path) => {
+        store.set('userPrefs.app.live-feed-location', path[0]);
+        const locationLabel = document.getElementsByClassName('sub-label livefeed');
+
+        for (let i = 0, len = locationLabel.length; i < len; i += 1) {
```
navdeepsinghkhalsa: How many places is it supposed to update? Is a `for` loop necessary?

tsingh777: At the moment 2 places (display tab and setting pane), when https://github.com/KhalisFoundation/sttm-desktop/issues/280 is done, it should be one place, can refactor once that change is in. 

for loop might be much but it gets the job done.

app.js
```diff
@@ -293,7 +293,8 @@ ipcMain.on('clear-apv', () => {
 });
 
 function createBroadcastFiles(arg) {
-  const userDataPath = electron.app.getPath('userData');
+  const liveFeedLocation = store.get('userPrefs.app.live-feed-location');
+  const userDataPath = (liveFeedLocation === 'default' || !liveFeedLocation) ? electron.app.getPath('desktop') : liveFeedLocation;
```
tsingh777: I hope `electron.app.getPath('desktop')` works for windows/linux as well. only tested on mac. 

cc. @navdeepsinghkhalsa

navdeepsinghkhalsa: It's officially documented, so they should be doing the legwork, here. I appreciated the find. Good work :)

PR #343, opened 2018-10-11
Fix 323 hover regression

navdeepsinghkhalsa:


navdeepsinghkhalsa:
@tsingh777 
This is what 4.1 has:
screen shot 2018-10-13 at 5 59 50 pm

This is what was introduced into 4.2 from fixing the hover bug on theme icons:
image 1

The above PR restores the hover from 4.1 while retaining your fix for theme hovers.

Issue #344, opened 2018-10-11
Look at when search should be disabled.


navdeepsinghkhalsa:
@tsingh777: _One concern I have [...] is when the download is happening, should we disable searching? ang search is still available when the initial db downloads and full searching features are available while the db updates. probably not a good idea for a user to attempt searching while we download/install db._

Though it shouldn't be a problem to search while a database is updating, the Ang search not being disabled during initial download is an issue and should be addressed.

It should just follow the same rules as the other input field and be disabled by default at search.js:36 and re-enabled at search.js:318

PR #346, opened 2018-10-16
Fix #336 Disable search during database download

### LINKED ISSUE #336: Unable to use app when DB is downloading
When launching the app for the first time, the app is unusable when the DB is downloading.

navdeepsinghkhalsa:


tsingh777:
I think we need to disable ang search as well when its downloading, both the initial and the update. 
load

navdeepsinghkhalsa:
@tsingh777 I agree. I created an issue for it here: https://github.com/KhalisFoundation/sttm-desktop/issues/344

tsingh777:
I still see issues with the download in this PR.

navdeepsinghkhalsa:
Nothing came up in my testing, Singh Jee. Could you please elaborate?

tsingh777:
While the DB is downloading, and a user enters something into the ang search, the progress bar shifts into the middle of the search bar (the gif above shows this.) Also in this scenario once the DB download is done, the placeholder becomes `undefined` (also in the gif).

if we aren't disabling the ang search in this PR, then it's fine no other issues to report.

navdeepsinghkhalsa:
Achhaa Jee. Thank you. I'll add the fix for #344 to the PR

PR #361, opened 2018-11-05
Only use Electron auto-updater if not published in an app store

navdeepsinghkhalsa:


app.js
```diff
@@ -7,6 +7,9 @@ const express = require('express');
 const fs = require('fs');
 const path = require('path');
 
+// Are we packaging for a platform's app store?
+const appstore = false;
```
tsingh777: When will this be true?

navdeepsinghkhalsa: For now, I will set it manually before starting the build, since the Mac App Store build has to be done locally anyway. If we give this responsibility to a CI server, then we can change it with a command from an NPM script.

PR #362, opened 2018-11-06
Custom background theme

inderpreetsingh:
screenshot 2018-11-06 at 4 37 47 pm
screenshot 2018-11-06 at 4 38 03 pm

Need to have a conversation with @saintsoldierx regarding this, before I make it final. 🙏


www/js/viewer.js
```diff
@@ -86,6 +86,20 @@ const castText = (text, isGurmukhi) => {
   castToReceiver();
 };
 
+const applyThemebg = () => {
+  if (prefs.app.themebg.url) {
+    $body.style.backgroundImage = `url(${prefs.app.themebg.url})`;
+  }
+
+  if (prefs.app.themebg.type === 'custom') {
+    $body.classList.add('show-overlay');
```
inderpreetsingh: Added overlay to prevent bad typography.

maneetpaul:
I would like to take this a step further and allow people to change the color of the individual elements on the slide as well (Gurmukhi color, translation color, transliteration color, etc). This was possible in STTM2 by creating a custom powerpoint.

ManjotS:
Agree. I think there is a separate ticket for fonts/colors. If not, it
should be a separate ticket as its a different thing to build.

saintsoldierx:
As mentioned Inderpreet Singh, let's try highlighting the text with contrasting opacity.

saintsoldierx:
image
image
@inderpreetsingh

inderpreetsingh:
screenshot 2018-12-03 at 8 21 27 pm
screenshot 2018-12-03 at 8 23 17 pm
screenshot 2018-12-03 at 8 23 52 pm
screenshot 2018-12-03 at 8 24 21 pm
screenshot 2018-12-03 at 8 24 43 pm

@saintsoldierx here is work in progress, It looks *okay* but it is heavily dependent upon the composition of the background image. So there is still a lot of theme + image combinations where text would be bit unreadable. We have to trust users to choose the right pictures, which I frankly am afraid to do :P Any other ideas on how to deal with text unreadability. I tried the glow effect, unfortunately, i lost those screenshots but results were not as good as the highlighted text.

What do you think about adding a gaussian blur to the background? or maybe we can just the trust the user.

maneetpaul:
Just an idea here, would this look better if instead of having each text element on the slide have a separate background overlay, we just have one giant slightly transparent overlay over the entire background image?

I’m not a fan of the individual overlays for each piece of text.

inderpreetsingh:
@maneetpaul veerji and I discussed yesterday, and he gave an idea that for now, custom background image should be treated like a separate theme. Where there's an overlay over background (light/dark) and the text on light background is dark and text on dark background is light. So we don't add custom background to existing themes but create a separate theme solely for purpose of accommodating custom background images. I think I will still provide options of dark and light overlay, but that would be a good enough start. 🙏

ManjotS:
isn’t it better to allow people to create n number of themes? gurdwaras
might want a normal sunday one, one for gurparabs, weddings etc

inderpreetsingh:
@ManjotS yes, that's a good idea. By creating a theme, do you also mean changing text colors and all?

Either way, I would strongly suggest to tackle it iteratively. For now, custom theme background supports only `light-theme`, `dark-theme`, `high-contrast` and `low-light`. 

screenshot 2018-12-14 at 3 29 46 pm

www/js/themes.json
```diff
@@ -26,6 +26,7 @@
     "transliteration-color": "#413b3d",
     "teeka-color": "#723c03",
     "background-image": "thumbnails/golden.jpg",
+    "background-image-full": "golden.jpg",
```
inderpreetsingh: Moved background image in json, to manipulate it via js instead of scss.

www/src/scss/_theme-a-new-day.scss
```diff
@@ -2,6 +2,10 @@ body.a-new-day {
   background-color: #07ccf3;
   background-image: url("../../assets/img/custom_backgrounds/clouds.jpg");
 
+  .color-overlay {
+    background-color: #07ccf3;
```
inderpreetsingh: added color overlays to all themes ( as i was trying to support all themes initially) but ended up using only light-theme, dark-theme, high-contrast and low-light.

www/src/scss/_theme-common.scss
```diff
@@ -0,0 +1,11 @@
+$highlightStregnth: 0.8;
```
inderpreetsingh: this is for highlighting single lines rather than adding an overlay on the whole background. Something that I am trying in sync with aman veerji. To have a general purpose custom background theme option that works with all the themes.

www/js/settings.js
```diff
@@ -319,6 +319,9 @@ module.exports = {
             });
             break;
 
+          case 'dynamic':
```
navdeepsinghkhalsa: What is the purpose of this, Singh Jee?

www/js/theme_editor.js
```diff
@@ -1,36 +1,16 @@
...
-    }
-  }
-  return defaultTheme;
-}; */
+const themesWithCustomBg = themes
+                            .filter(theme => theme.type === 'COLOR' || theme.type === 'SPECIAL')
+                            .map(theme => theme.key);
 
-const swatchFactory = themeInstance =>
+const swatchFactory = (themeInstance, isItCustom) =>
```
navdeepsinghkhalsa: Convention is usually to omit the `It` and just have `isCustom`
Also, it might be best to assign a default value of `false`

www/js/theme_editor.js
```diff
@@ -68,70 +52,61 @@ const swatchFactory = themeInstance =>
...
+        type: 'file',
+        accept: 'image/*',
+        onchange: (evt) => {
+          const curTheme = store.getUserPref('app.theme');
+          if (!themesWithCustomBg.includes(curTheme)) {
+            store.setUserPref('app.theme', themesWithCustomBg[0]);
+          }
+          store.setUserPref('app.themebg', {
+            type: 'custom',
+            url: evt.target.files[0].path.replace(/(\s)/g, '\\ '),
```
navdeepsinghkhalsa: Could we please run an optimizer on the image and make a copy of it in the userData directory and use that? https://www.npmjs.com/package/imagemin

www/src/scss/_theme-common.scss
navdeepsinghkhalsa: typo: Stregnth

navdeepsinghkhalsa: And it doesn't seem to be used anywhere, or am I missing it? :)

www/src/scss/_theme-a-new-day.scss
navdeepsinghkhalsa: Should we remove, as the `color-overlay` class seems to not be in use?

inderpreetsingh: it is used in the themes that currently support custom background. Initially, it was for all themes, so colour overlay existed everywhere. I can change it for now and add it later when we add support for more themes.

www/src/scss/_theme-common.scss
inderpreetsingh: It was used in supporting all the themes (when we were trying to use a different way to highlight the verses). I will remove it from this PR and keep it separate for the support of more themes.

navdeepsinghkhalsa:
I'm a bit uncertain with having duplicate theme swatches for the themes that support custom backgrounds. Is there a better way to present this? Maybe an icon indicator for those that support it?

inderpreetsingh:
@navdeepsinghkhalsa veerji, when you click on say "light theme" , "dark theme" from the normal swatches area, it removes the background. 

I can shift logic to always add the background image even when shifting themes and add another button named `remove background image` in case user wants say light theme without the bg image.

maneetpaul:
I think we are over complicating this a bit. Can we just have **ONE** option at the bottom of the list for Custom Background? And have it be the light theme? Don't even call it "Light". Just have it say "Custom".

We can revisit this in a future Sprint when we want to add additional customizations.

navdeepsinghkhalsa:
I would agree with either the remove button or a single theme for custom background, which later gives even more controls.

inderpreetsingh:
I will go with the only light theme ⚪️ as the first iteration, with light ⚪️ , dark ⚫️ , high contrast and low light as second iteration and `all themes` as the third iteration. In second and third iteration (separate PRs) I will add the remove background interface. 

We need light and dark themes because some images are too dark to look good under a white theme (say a night view of gurudwara at bandi chorr divas).

www/js/theme_editor.js
```diff
@@ -1,36 +1,26 @@
...
-        modal: true,
-      }).show();
-      return defaultTheme;
-    }
-  }
-  return defaultTheme;
-}; */
+const themesWithCustomBg = themes
+                            .filter(theme => theme.type === 'COLOR' || theme.type === 'SPECIAL')
+                            .map(theme => theme.key);
```
inderpreetsingh: This is not properly used yet, but will be used (and was inteded to be used) when supporting more than one theme.

PR #364, opened 2018-11-13
Fix 356 detach viewer from db

### LINKED ISSUE #356: 4.2 causes app to become unusable on windows 7
fb961531-e6fd-4853-94fa-682fe785c844

navdeepsinghkhalsa:
Fix #356 Send rows to viewer instead of doing a separate DB lookup inside viewer
Separate infinite scroll from APV
Fix phantom scroll on lists when using keyboard arrows for navigating
Fix result list height, which was hiding the last result


CHANGELOG.md
```diff
@@ -1,8 +1,5 @@
 ## SikhiToTheMax Desktop App
 
-### 4.2.2 - _181106_
```
tsingh777: let remember to update this before releasing.

www/js/search.js
```diff
@@ -249,8 +253,8 @@ document.body.addEventListener('click', e => {
 });
 
 function akhandPaatt() {
-  module.exports.$shabad.innerHTML = '';
-  global.controller.clearAPV();
+  // module.exports.$shabad.innerHTML = '';
```
tsingh777: add a TODO maybe? not a blocker.

PR #371, opened 2018-11-25
 adds asterisk as wild  char for realm search

inderpreetsingh:
screenshot 2018-11-25 at 5 38 03 pm
screenshot 2018-11-25 at 5 39 02 pm
screenshot 2018-11-25 at 5 39 45 pm

I am currently setting one `*` for one letter (one word). It felt more natural to me (keeping in mind our users), should we make one asterisk mean any number of characters? 

In that case `*jtmv` would mean any number of characters before `jtmv`. @ManjotS @saintsoldierx

### DIFF
--- a/www/js/banidb/sqlite-search.js
+++ b/www/js/banidb/sqlite-search.js
@@ -43,6 +43,8 @@ const query = (searchQuery, searchType, searchSource) => (
     let dbQuery = '';
     let searchCol = '';
     let condition = '';
+    let isWildChar = false;
+
     // Sanitize query
     const saniQuery = searchQuery.trim().replace("'", "\\'");
     // default source for ang search to GURU_GRANTH_SAHIB
@@ -58,7 +60,12 @@ const query = (searchQuery, searchType, searchSource) => (
           if (charCode < 100) {
             charCode = `0${charCode}`;
           }
-          dbQuery += `,${charCode}`;
+          if (charCode === '042') {
+            isWildChar = true;
+            dbQuery += ',___';
+          } else {
+            dbQuery += `,${charCode}`;
+          }
         }
 
         // Replace kh with kh pair bindi
@@ -68,7 +75,7 @@ const query = (searchQuery, searchType, searchSource) => (
         }
 
         // Use LIKE if anywhere, otherwise use operators
-        if (searchType === CONSTS.SEARCH_TYPES.FIRST_LETTERS_ANYWHERE) {
+        if ((searchType === CONSTS.SEARCH_TYPES.FIRST_LETTERS_ANYWHERE) || isWildChar) {
           condition = `${searchCol} LIKE '%${dbQuery}%'`;
           if (replaced) {
             condition += ` OR ${searchCol} LIKE '%${replaced}%'`;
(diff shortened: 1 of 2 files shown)


maneetpaul:
@inderpreetsingh - is there any slowdown or lag when using the `*` to query multiple words instead of just one?

If not, there is a huge benefit to have the `*` check for multiple words. There have been multiple situations where I was doing a search and was unable to understand one of the words the keertanyee was singing. After finding the shabad, it turned out that one of the words was actually two separate words. Being able to use an `*` when searching would have solved that.

ManjotS:
This is especially useful in the case of sat nam vs satnam or gur sikh or
gursikh

saintsoldierx:
@inderpreetsingh i am good with the * - what other input do you need from me?

tarunsingh5:
perhaps a separate char for more than one char per wildcard? like * = 1 char but # = 1+ char?

maneetpaul:
We should stick with one character. The point of this is sometimes you
don’t know when there are more than one word. So you wouldn’t know to use
the other char.

tarunsingh5:
hmm but im pretty sure there's going to be a speed tradeoff - @navdeepsinghkhalsa thoughts?

ManjotS:
theres a reason search engines were designed with certain characters..
these translate well to databases.

* translates well to mysql/mariadb % which is one or more characters. so
this doesn’t really need to be discussed further unless realm limits it to
1 char.

The performance difference of one char wild vs multiple chars is almost
negligible, its still an unknown.

inderpreetsingh:
In realm * is for multiple characters and ? is for one character. 

I will move with multiple characters ahead.

navdeepsinghkhalsa:
@inderpreetsingh Could you please also add the SQLite version in this PR?

inderpreetsingh:
@navdeepsinghkhalsa veerji apologies, I did not understand your question? Do you mean that by putting an asterisk before the search query "from start" can be turned into "from anywhere"?

www/js/banidb/sqlite-search.js
```diff
@@ -68,7 +75,7 @@ const query = (searchQuery, searchType, searchSource) => (
         }
 
         // Use LIKE if anywhere, otherwise use operators
-        if (searchType === CONSTS.SEARCH_TYPES.FIRST_LETTERS_ANYWHERE) {
+        if ((searchType === CONSTS.SEARCH_TYPES.FIRST_LETTERS_ANYWHERE) || isWildChar) {
```
navdeepsinghkhalsa: By coupling `isWildChar` with `searchType` being first-letter-anywhere, wouldn't this `AGn*` be the same as `*AGn*`?

tarunsingh5: I was thinking about this a bit on the drive this morning and I think @navdeepsinghkhalsa is correct. That doing a a wildcard at the start of a search string for First Letter Start is effectively doing first letter anywhere - the only time the they are different is when it is not the first letter that is the wildcard

ManjotS:
My thoughts are its fine that adding a star at the beginning does anywhere,
we should just have it go to that code path and auto update the drop down
to anywhere when it happens. If they backspace, just go to previous drop
down setting.

www/js/banidb/sqlite-search.js
inderpreetsingh: @navdeepsinghkhalsa I have given various cases in slack, but to answer your this specific question, yes in first letter annywhere, `AGn*` will effectively be `*AGn*`.

Issue #245, opened 2018-08-11
"Escape" key should show blank slide


maneetpaul:
Great to use between keertanyees.

PR #372, opened 2018-11-25
Cmd +e / Ctrl + e shows empty slide

inderpreetsingh:
Earlier clicking on escape closed settings that code was commented. However for issue #245 we want to show an empty slide when clicked on escape button.

### DIFF
--- a/www/js/index.js
+++ b/www/js/index.js
@@ -12,6 +12,7 @@ function escKey() {
   /* if (settings.$settings.classList.contains('animated')) {
     settings.closeSettings();
   } */
+  global.controller.sendText('');
 }
 
 function highlightLine(newLine) {
--- a/www/js/index.js
+++ b/www/js/index.js
@@ -12,6 +12,11 @@ function escKey() {
   /* if (settings.$settings.classList.contains('animated')) {
     settings.closeSettings();
   } */
+
+  global.controller.sendText('');
+}
+
+function hideSlide() {
   // show Empty Slide
   global.controller.sendText('');
 }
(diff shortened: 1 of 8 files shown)


www/js/index.js
```diff
@@ -12,6 +12,8 @@ function escKey() {
   /* if (settings.$settings.classList.contains('animated')) {
     settings.closeSettings();
   } */
+  // show Empty Slide
+  global.controller.sendText('');
```
navdeepsinghkhalsa: This is being bound with bindGlobal, which will tigger this even if `esc` is pressed while the cursor is in a text field. That doesn't seem intuitive. @maneetpaul @saintsoldierx Do you have any thoughts on what criteria need to be met for the `esc` key to trigger an empty slide?

inderpreetsingh: I agree! Its bad UX to repurpose keystrokes like that. I think we should do something like `cmd + e` or `ctrl+e` for this.

saintsoldierx:
Having discussed with Inder Singh, Esc key can present other issues when using STTM - so let's use CMD + E for now as hotkey. Also, this should toggle the secondary screen on and off - not just show a blank screen.

maneetpaul:
@saintsoldierx can you clarify what you mean by “turn off the second display”? My original intention of this ticket was a quick way to clear all the text off the slide (while retaining the background of the theme). I don’t want to change the scope to actually turn off the entire display.

inderpreetsingh:
@saintsoldierx I agree with maneetpaul veerji we should have a separate functionality for closing secondary window. we talked about having a close icon on viewer preview, clicking on this icon would dim the viewer preview (showing it as its disabled) and close the secondary viewer window.

inderpreetsingh:
Changed from esc to cmd+e

ManjotS:
what is windows equivalent?

inderpreetsingh:
@ManjotS it's `ctrl + e`

www/js/index.js
navdeepsinghkhalsa: This was later decided to be cmd/ctrl+e, so I will click "resolve" for this conversation

PR #373, opened 2018-11-25
adds new toggle for presenter view

### LINKED ISSUE #295: add presenter view toggle button to current shabad (single monitor only)
maybe  to title or bottom right. This is useful in single monitor if you want to do search in presenter and then quickly switch to full slide.

perhaps @saintsoldierx can provide design and icon location guidance

inderpreetsingh:
ezgif-3-981faf951bcb

It fixes #295 

I have tested it with chromecast and single monitor. Yet to test it with a dual monitor setup.

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -167,12 +167,13 @@
       <div id="footer">
         <div class="menu-group">
           <div class="menu-group-left"></div>
+          <div class="presenter-switch"> </div>
           <!-- <div class="menu-group-right">
-            <ul class="menu-bar">
+             <ul class="menu-bar">
               <li>
                 <a class="action" data-action="settings.openSettings"><span class="small"><i class="fa fa-fw fa-cog"></i></span><span class="medium">Settings</span></a>
               </li>
-            </ul>
+            </ul> 
           </div> -->
         </div>
       </div>
--- a/www/index.html
+++ b/www/index.html
@@ -168,13 +168,6 @@
         <div class="menu-group">
           <div class="menu-group-left"></div>
           <div class="presenter-switch"> </div>
-          <!-- <div class="menu-group-right">
-             <ul class="menu-bar">
-              <li>
-                <a class="action" data-action="settings.openSettings"><span class="small"><i class="fa fa-fw fa-cog"></i></span><span class="medium">Settings</span></a>
-              </li>
-            </ul> 
-          </div> -->
         </div>
       </div>
     </div>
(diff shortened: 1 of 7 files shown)


navdeepsinghkhalsa:
@saintsoldierx Could you please sign off on design before code review?

inderpreetsingh:
@navdeepsinghkhalsa took a verbal sign off from aman veerji over call! 🙈 @saintsoldierx

www/index.html
```diff
@@ -167,12 +167,13 @@
       <div id="footer">
         <div class="menu-group">
           <div class="menu-group-left"></div>
+          <div class="presenter-switch"> </div>
```
tsingh777: should we remove the one in the settings eventually? doesn't have to be in this pr.

www/js/controller.js
```diff
@@ -386,7 +386,12 @@ function checkPresenterView() {
     document.body.classList.add('presenter-view');
     document.body.classList.remove('home');
     document.body.classList.add('scale-viewer');
+  } else {
+    document.body.classList.remove('presenter-view');
+    document.body.classList.add('home');
+    document.body.classList.remove('scale-viewer');
   }
+  document.querySelector('#presenter-view-toggle').checked = store.getUserPref('app.layout.presenter-view');
```
tsingh777: could refactor this function to less lines with something like
```
function checkPresenterView() {
  const inPresenterView = store.getUserPref('app.layout.presenter-view');
  const classList = document.body.classList;

  classList.toggle('presenter-view', inPresenterView);
  classList.toggle('home', !inPresenterView);
  classList.toggle('scale-viewer', inPresenterView);

  document.querySelector('#presenter-view-toggle').checked = inPresenterView;
}
```

your call.

www/index.html
```diff
@@ -167,12 +167,13 @@
...
           <!-- <div class="menu-group-right">
-            <ul class="menu-bar">
+             <ul class="menu-bar">
```
tsingh777: commented code should be removed eventually.

www/index.html
inderpreetsingh: Yeah need to discuss that with aman veerji. All these options are now at two places (specially display options).

www/index.html
inderpreetsingh: Yeah, I agree. I will remove it.

maneetpaul:
Tracking this in #295 - are we all set with this PR?

inderpreetsingh:
Yeah, its ready from my side. @maneetpaul

PR #375, opened 2018-11-26
[WIP] Bisrams

inderpreetsingh:
screenshot 2018-11-26 at 5 06 22 pm

screenshot 2018-11-26 at 5 06 13 pm

### DIFF
--- a/www/js/defaults.json
+++ b/www/js/defaults.json
@@ -35,6 +35,7 @@
         "disable-akhandpaatt": false,
         "akhandpaatt": false,
         "larivaar": false,
+        "display-bisrams": true,
         "left-align": false
       }
     }
--- a/www/js/settings.json
+++ b/www/js/settings.json
@@ -68,6 +68,7 @@
             "subLabel": "(Google ChromeCast Unavailable)"
           },
           "larivaar": "Larivaar",
+          "display-bisrams": "Bisrams",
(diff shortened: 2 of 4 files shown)


inderpreetsingh:
I have used google orange and google green as default. We need to define separate colors for separate themes. _discussion in channel `#visraams_discussion`

ManjotS:
those colors are really nice in the screenshot

tarunsingh5:
also lets update spellings to be "Vishraams"

inderpreetsingh:
> also lets update spellings to be "Vishraams"

It's bisram in db, so I used that.

tarunsingh5:
where are we on this?

inderpreetsingh:
@tarunsingh5  waiting for colors for themes. However @maneetpaul veerji gave me an idea yesterday. I think to begin with we should just support light and dark theme, where there's no confusion about colours. And add support for themes as next step.

inderpreetsingh:
@navdeepsinghkhalsa veerji, aah! I wasn't aware of that. I am not well versed with santhya, so it seems yamkis are very rare, right?

navdeepsinghkhalsa:
@inderpreetsingh Thank you for making the update! In fact, they're not that rare. Our data is incomplete and classifies many Yamkis as Visraams. This will be updated in the future.

Issue #15, opened 2018-01-12
Pressing a letter key should go to the next line that starts with that letter and then loop around


tarunsingh5:
Pressing a letter key should go to the next line that starts with that letter and then loop around

inderpreetsingh:
Interesting! I like that 👏

PR #376, opened 2018-12-05
Pressing a letter should go to that panktee in current shabad panel.

inderpreetsingh:
Solves #15 

# click below image to see video

[Video for key jumps](http://www.youtube.com/watch?v=2L6_JwAXwCM)

* You have to click on current shabad panel, focus it before you can start using the keys.

### DIFF
--- a/www/js/index.js
+++ b/www/js/index.js
@@ -84,12 +84,12 @@ function findLine(e) {
   const pos = search.currentShabad.indexOf(search.currentLine);
 
   // Rotate the array based on current shabad
-  const panktees = Array.from(search.$shabad.getElementsByClassName('panktee'));
+  const panktees = [...search.$shabad.getElementsByClassName('panktee')];
   const pankteesBeforePos = panktees.splice(0, pos + 1);
-  const pankteesRotated = panktees.concat(pankteesBeforePos);
+  const pankteesRotated = [...panktees, ...pankteesBeforePos];
 
   const lineFound = pankteesRotated.find((panktee) => {
-    const pankteeText = panktee.querySelector('.main-letters').innerText;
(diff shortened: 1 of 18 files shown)


www/js/index.js
```diff
@@ -71,15 +76,43 @@ function nextLine(e) {
...
 
+function findLine(e) {
+  e.preventDefault();
+  const filterKey = e.key;
+
+  // Find position of current line in shabad
+  const pos = search.currentShabad.indexOf(search.currentLine);
+
+  // Rotate the array based on current shabad
+  const panktees = Array.from(search.$shabad.getElementsByClassName('panktee'));
```
navdeepsinghkhalsa: `[...search.$shabad.getElementsByClassName('panktee')]`

www/js/search.js
```diff
@@ -671,6 +671,10 @@ module.exports = {
             h('i.fa.fa-fw.fa-home'),
             ' ',
             item.Gurmukhi,
+            h(
+              'span.hidden.main-letters',
+              item.MainLetters,
+            ),
```
navdeepsinghkhalsa: Could this be a data attribute instead of adding them to the DOM? That way you can avoid a costly `querySelector` and just get the value from the attribute itself.

www/js/index.js
```diff
@@ -71,15 +76,43 @@ function nextLine(e) {
...
+  const pankteesBeforePos = panktees.splice(0, pos + 1);
+  const pankteesRotated = panktees.concat(pankteesBeforePos);
```
navdeepsinghkhalsa: `[ ...panktees, ...pankteesBeforePos ];`

PR #379, opened 2018-12-12
Announcement formatting

### LINKED ISSUE #269: Announcements feature needs to reflect proper formatting and should be centered
screen shot 2018-08-16 at 9 59 01 pm

inderpreetsingh:
screenshot 2018-12-12 at 6 32 11 pm

This fixes #269 , adds formatting to the announcement. To have even more formatting we can have a WYSIWYG or markdown editor, but I think this is good enough to start with. We can track analytics on this and then see how much people use it and decide priority accordingly. Also, the announcement is vertically in the centre now. The centring does not work in chromecast but the formatting is still preserved.

### DIFF
--- a/www/js/menu.js
+++ b/www/js/menu.js
@@ -251,28 +251,31 @@ const announcementSlideButton = h(
               name: 'announcement-language',
               type: 'checkbox',
               onclick: () => {
-                const $announcementText = document.querySelector('.announcement-text');
-                $announcementText.classList.toggle('gurmukhi');
                 const isGurmukhi = document.querySelector('#announcement-language').checked;
-                $announcementText.value = '';
-                $announcementText.placeholder = isGurmukhi ? 'GoSxw ie`Qy ilKo ...' : 'Add announcement text here ..';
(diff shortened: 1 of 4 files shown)


ManjotS:
This is definitely a security issue. If we are doing this it should only
allow a pre approved list of tags such as i,b,u etc

ManjotS:
or just put in a lightweight wysiwyg

maneetpaul:
I agree with the above comments on limiting the HTML tags that can be used to be basic/pre-approved ones for formatting purposes. 

@inderpreetsingh can we adjust accordingly?

inderpreetsingh:
used `sanitize-html` to sanitize the html and allowed following tags `['b', 'i', 'em', 'u', 'pre', 'strong', 'div', 'code', 'br', 'p', 'ul', 'li', 'ol'],`

navdeepsinghkhalsa:
@inderpreetsingh Thank you for updating this! Could you also please run the sanitizer on keydown (including when pasting)?

PR #380, opened 2018-12-19
Update Event tracking and setting default

tsingh777:
Enable tracking by default. 

This changes the event sending from each search to shabad selection.

### DIFF
--- a/www/js/search.js
+++ b/www/js/search.js
@@ -513,7 +513,6 @@ module.exports = {
     } else {
       this.$results.innerHTML = '';
     }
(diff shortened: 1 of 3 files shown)


www/js/search.js
```diff
@@ -593,6 +592,14 @@ module.exports = {
     // scroll the session block to the top to see the highlighted line
     this.$sessionContainer.scrollTop = 0;
     this.navPage('shabad');
+    const searchType = this.searchType;
```
navdeepsinghkhalsa: Can object destructuring be used with `this`?

Issue #386, opened 2018-12-21
Error when uploading custom BG


maneetpaul:
What are the image file requirements for it to work? Can we make the error message more user-friendly so they know why it doesn't work?

<img width="322" alt="screen shot 2018-12-21 at 12 33 26 pm" src="https://user-images.githubusercontent.com/10658257/50355462-c42cbc00-051c-11e9-861a-6db3d8856052.png">

PR #390, opened 2018-12-21
Fix uploading a custom background images

tsingh777:
Fixes https://github.com/KhalisFoundation/sttm-desktop/issues/386


navdeepsinghkhalsa:
Could you please test the difference in optimization efficiency of imagemin on its own vs with the plugins?

tsingh777:
here's the metrics:

original file size and type |  imagemin only | with pluggins 
-|-|-
| jpeg  2.4mb| 2.4mb | 2.3mb
| jpeg 590kb| 590kb | 5.85 kb
|png 1.3mb| 1.3mb | 375kb
|png 366kb| 366kb |  101kb | 

looks like its not that efficient for `jpeg` but **very** efficient for `png`s. 

Would require more debugging to see why the pluggins don't work in the built version of the app. 

1 alternative option would be to clear out the `user_backgrounds` folder every time they "upload" and only "save" a single image in that folder, that way we don't duplicate multiple images on their machine. 

Also the file input seems to be ignoring the `accept` attribute and allowing all files to be selected, not sure if that's a limitation with hyperscript. 

```
h('input.file-input#themebg-upload',
      {
        type: 'file',
        accept: 'image/png, image/jpeg',
```

www/js/theme_editor.js
```diff
@@ -81,31 +88,26 @@ const imageInput = () =>
           // if (!themesWithCustomBg.includes(curTheme)) {
           store.setUserPref('app.theme', themesWithCustomBg[0]);
           // }
-          try {
-            const userBackgroundsPath = `${userDataPath}/user_backgrounds`;
+          const userBackgroundsPath = `${userDataPath}/user_backgrounds`;
 
+          try {
```
tsingh777: added another try so we don't swallow all errors in one block.

PR #403, opened 2018-12-25
changes the focus pattern

inderpreetsingh:
screenshot 2018-12-25 at 5 10 41 pm


www/src/scss/styles.scss
```diff
@@ -164,6 +164,19 @@ body.show-overlay {
...
+    .current-shabad-header {
+      border-bottom: 2px solid transparent;
+      transition: 0.5s border ease;
+    }
+
+    &:focus {
+      outline: 0;
+
+      .current-shabad-header {
+        border-bottom: 2px solid #01579b;
```
navdeepsinghkhalsa: Could just use `border-bottom-color` :)

PR #432, opened 2019-01-11
Hide vishraams

inderpreetsingh:
Hide vishraams for now.

### DIFF
--- a/www/js/defaults.json
+++ b/www/js/defaults.json
@@ -39,9 +39,6 @@
         "disable-akhandpaatt": false,
         "akhandpaatt": false,
         "larivaar": false,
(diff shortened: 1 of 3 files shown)


www/js/defaults.json
```diff
@@ -39,10 +39,8 @@
         "disable-akhandpaatt": false,
         "akhandpaatt": false,
         "larivaar": false,
-        "display-bisrams": true,
```
tsingh777: you should be able to make this false instead of deleting? up to you.

navdeepsinghkhalsa:
@inderpreetsingh Jee, could you please resolve conflicts so we can get a truer understanding of the diff?

PR #438, opened 2019-01-12
Adds custom bg fix to windows

inderpreetsingh:
custom background stopped working on windows. Converted backslash to forward slash and it worked. Also had to use `path.resolve` for combining paths. Earlier we were just concatenating strings.

### DIFF
--- a/www/js/search.js
+++ b/www/js/search.js
@@ -601,6 +601,7 @@ module.exports = {
       searchValue = this.$search.value;
     }
     analytics.trackEvent('search', CONSTS.SEARCH_TEXTS[searchType], searchValue);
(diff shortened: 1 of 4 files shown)


inderpreetsingh:
@navdeepsinghkhalsa @tsingh777  is there a way to use node `path` module or something like that to convert windows based file paths into css compatible strings?

I am using regex but i am not sure if this is the right way to go. 

http://cwestblog.com/2017/06/16/css-local-file-system-paths-on-windows/

navdeepsinghkhalsa:
@inderpreetsingh You're right - Our version of Electron doesn't come with v10

PR #446, opened 2019-01-15
Revert GurmukhiBisram usage

navdeepsinghkhalsa:
Commenting out until database is updated with better data

### DIFF
--- a/www/js/viewer.js
+++ b/www/js/viewer.js
@@ -218,26 +218,26 @@ const createCards = (rows, LineID) => {
   Object.keys(rows).forEach((key) => {
     row = rows[key];
     lines.push(row.ID);
-    const taggedGurmukhi = row.Gurmukhi.split(' ');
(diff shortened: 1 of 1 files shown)


www/js/viewer.js
```diff
@@ -218,7 +218,8 @@ const createCards = (rows, LineID) => {
   Object.keys(rows).forEach((key) => {
     row = rows[key];
     lines.push(row.ID);
-    const gurmukhiShabads = row.GurmukhiBisram.split(' ');
+    const taggedGurmukhi = row.Gurmukhi.split(' ');
```
tsingh777: Are there any `]` characters in this column? In which case we would need that loop for the first condition around line 224?

navdeepsinghkhalsa: Updated to just revert the changes from here: https://github.com/KhalisFoundation/sttm-desktop/commit/31bac4747c85b05669e80017e61e96f2c88458c2#diff-df6fefbebc662cf1cd3f5236a8ae670d

PR #455, opened 2019-01-21
downloads and keeps the help images local

### LINKED ISSUE #408: Move help from web to desktop
Help section should be local to the app

inderpreetsingh:
This fixes #408


tsingh777:
LGTM. @maneetpaul should we add the videos as well? they don't work unless the user is connected to the internet.

maneetpaul:
If we do videos local too wouldn’t that increase the file size of the app by a lot?

tsingh777:
Size would depend on the quality I suppose. 

Should we show something in addition that doesn't rely on internet?

tarunsingh5:
@tsingh777 maybe we can make some gifs?

maneetpaul:
GIFs would be best. Let’s leave the embedded YouTube for now and we’ll add GIFs when I update the help next time

tsingh777:
GIFs won't have sound but yeah, we can add them later. I was thinking of leaving a caption under the videos that has a link or something. Can track that separately @maneetpaul.

Issue #382, opened 2018-12-19
Consolidate Duplicate Settings


maneetpaul:
Find a central place to keep all display settings rather than two separate locations.

maneetpaul:
Thanks @saintsoldierx 

Currently, all the tabs are under the history section. These need to be moved to the bottom bar rather than nested under History. 

This will allow us to remove the preferences menu completely and allow easy access to all the settings in an intuitive way.

maneetpaul:
@saintsoldierx what are your thoughts on renaming the "Display" tab to "Settings" instead? The icon is already reflective of the word "Settings" so it would just be a textual change.

PR #467, opened 2019-02-02
Rearrange settings

inderpreetsingh:
https://youtu.be/nfc8uvB4HCg

I haven't removed settings from the original settings panel. I was thinking maybe we can remove it gradually when analytics data shows and confirms that people are used to the new interface. I don't want to give a design shock. 

I also think we should clean HTML for this part of the app (use hyperscript + json files to generate html). I will create a separate issue for it. Maybe we can scope it along with the upcoming CSS cleanup.  🙏 

Regarding not removing setttings, me and maneetpaul veerji discussed and came to following conclusion.
* [x] Keep the settings icon in presenter view. When you click it, it will open the display tab on the bottom right
* [x] Adding a focus effect of some sort to draw attention that it is on the bottom right now
* [x]  Also changing it so when you go to preferences on the top menu bar (the one I showed you) it also opens the tab on the bottom right
* [x] Moving all the display options to the top, and moving collect statistics to the bottom. Removing the  button for Presenter View.
* [x] adding an Analytic tag for that presenter view settings button


maneetpaul:
Fixes #382

tarunsingh5:
@tsingh777 @navdeepsinghkhalsa ptal

www/js/footer-left.json
```diff
@@ -1,14 +1,37 @@
...
+    "icon": "paint-brush",
+    "type": "tab-nav"
+  },
+  "insert": {
+    "label": "Insert",
+    "icon": "desktop",
+    "type": "tab-nav"
+  },
+  "display": {
+    "label": "Display",
```
tsingh777: Should it still be called Display? its more like settings or preferences now since it has all the other options. @maneetpaul

www/js/menu.js
```diff
@@ -329,6 +330,23 @@ module.exports = {
 
   showNotificationsModal,
 
+  showDisplayTab(fromMainMenu) {
+    search.activateNavLink('display', true);
+    search.activateNavPage('session', { id: 'display', label: 'Display' });
+
+    const isPresenterView = document.body.classList.contains('presenter-view');
+    let settingsClickSource = isPresenterView ? 'from_presenter_view' : 'not_from_presenter_view';
+    if (fromMainMenu) settingsClickSource += '_main_menu';
```
tsingh777: i don't think the scenario should appended into one string. try to separate it out so the analytics can be filtered on better. @maneetpaul can provide some input on what would be easiest to report on.
```
 ------------------------------------------
   * eventCategory  | text |  yes     | Typically the object that was interacted with (e.g. 'Video')
   * eventAction    | text |  yes     | The type of interaction (e.g. 'play')
   * eventLabel     | text |  no      | Useful for categorizing events (e.g. 'Fall Campaign')
   * eventValue     | int  |  no      | A numeric value associated with the event (e.g. 42)
```

www/js/menu.js
```diff
@@ -329,6 +330,23 @@ module.exports = {
...
+    analytics.trackEvent('newSettings', settingsClickSource);
+
+    const sessionPage = document.querySelector('#session-page');
+    sessionPage.classList.add('bounce-animate');
+    sessionPage.offsetWidth = undefined; // reflow to restart animation
```
tsingh777: what does this do? seems like a hack.

www/js/search.js
```diff
@@ -216,20 +216,40 @@ const searchOptions = h(
...
   navPageLinks.push(
     h(
       'li',
       h(
-        `a#${id}-pageLink`,
+        `a#${id}-pageLink.${navType}`,
         {
-          onclick: () => module.exports.navPage(id),
+          onclick: () => {
+            if (navType === 'pane-nav') {
```
tsingh777: can you add some comments here? little confusing to follow.

www/js/settings.json
```diff
@@ -1,36 +1,27 @@
...
         "type": "switch",
         "options": {
-          "presenter-view": "Presenter View"
+          "akhandpaatt": {
+            "label": "Akhand Paatth View (Beta)",
+            "subLabel": "(Google ChromeCast Unavailable)"
+          },
+          "larivaar": "Larivaar",
+          "left-align": "Left-Align",
+          "autoplay": "Auto Play",
```
tsingh777: this is a nit but can this option be closer to the autoplay delay slider?

www/js/menu.js
```diff
@@ -304,7 +305,7 @@ module.exports = {
     const $preferencesOpen = document.querySelectorAll('.preferences-open');
     $preferencesOpen.forEach(($menuToggle) => {
       $menuToggle.appendChild(menuButton.cloneNode(true));
-      $menuToggle.addEventListener('click', () => { module.exports.toggleMenu('#menu-page'); });
+      $menuToggle.addEventListener('click', module.exports.showDisplayTab);
```
tsingh777: > Keep the settings icon in presenter view. When you click it, it will open the display tab on the bottom right

do we need the settings icon in presenter view? its pretty far away mouse wise from what it opens in presenter view. it has some benefits in non-presenter view.

maneetpaul: Me and @inderpreetsingh had a discussion on this. We are unsure if people are using this button to access the settings. So rather than getting rid of it completely and confuse people, we are leaving it in place for now. However yes, that is the plan, to eventually remove this button as it is no longer needed in Presenter View

www/js/menu.js
maneetpaul: I am assuming this is for the analytics tag for the settings button. What I need to know: 

- How often the Settings button in the top left pane of Presenter View is clicked
- How often people click the settings button via the File > Preferences menu

www/js/footer-left.json
maneetpaul: I agree that this should be changed. I think "Settings" would be more appropriate.

inderpreetsingh: Okay.

www/js/menu.js
inderpreetsingh: It is used to restart CSS animation on click. You are right, it's a bit of a hack. I couldn't find a nice another way to restart CSS animations and I didn't want to use setTimeout just for the sake of this.

www/js/settings.json
inderpreetsingh: Okay

www/js/menu.js
inderpreetsingh: I am not happy with the UI we have at the moment either. I especially don't like how menu behaves in non-presenter view. However, that's all base don opinions, I would rather use real-world data to decide how to move further.

maneetpaul: I would like to play around with this once we have a new alpha is released with this. Let's discuss after that.

inderpreetsingh: Okay! I will push a quick separate PR and a video for what I think is the ideal solution. 🙏

inderpreetsingh:
@tsingh777 @maneetpaul I have implemented your suggestions. Please take a look. 🙏

PR #485, opened 2019-02-12
fixes launch error in mac

### LINKED ISSUE #483: Error message upon launch for Mac users
From @navdeepsinghkhalsa: I'm not sure at what point this started. It's happening because the `user-preferences.json` file doesn't exist, so we should check to see if the file exists before opening, and create a blank file if it doesn't.

This is in `www/js/store.js`
https://github.com/KhalisFoundation/sttm-desktop/blob/dev/www/js/store.js#L13
Just have to update this to `fs.readFileSync(filePath, { flag: 'w' })`

screen shot 2019-01-22 at 9 37 32 am

inderpreetsingh:
fixes #483

### DIFF
--- a/www/js/store.js
+++ b/www/js/store.js
@@ -10,7 +10,7 @@ function parseDataFile(filePath, defaults) {
   // which will be the case on the first application run.
   // `fs.readFileSync` will return a JSON string which we then parse into a Javascript object
   try {
-    return JSON.parse(fs.readFileSync(filePath, { flag: 'w' }));
+    return JSON.parse(fs.readFileSync(filePath, { flag: 'w+' }));
   } catch (error) {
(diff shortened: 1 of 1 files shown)


www/js/store.js
```diff
@@ -10,7 +10,7 @@ function parseDataFile(filePath, defaults) {
   // which will be the case on the first application run.
   // `fs.readFileSync` will return a JSON string which we then parse into a Javascript object
   try {
-    return JSON.parse(fs.readFileSync(filePath));
+    return JSON.parse(fs.readFileSync(filePath, { flag: 'w' }));
```
tsingh777: is `w` the correct flag? seems like `w+` is what we need according to the docs(https://nodejs.org/api/fs.html#fs_file_system_flags) but i could be wrong.

PR #493, opened 2019-02-13
Shortcut tray

### LINKED ISSUE #429: Add Shortcut Tray
To include:

- Anand Sahib + Salok 
- Waheguru Slide
- Blank Slide

This will only be available in Presenter View. It should have a button to allow people to collapse/minimize it in case they want to use the space for the History section instead. 

image 4

inderpreetsingh:
https://youtu.be/tSyYnr_syyQ

I have tried to make it as much like the mockup given as possible but I have following hiccups/concerns. Most of them are personal opinions, so feel free to override them.

* The current colours and sizes take more attention then I would normally prefer. (compared to rest of interface)
* We use border-radius on all our theme tiles, I would like to have that consistency.
* Background to text contrast ratio is arguable. It scores bad on WCAG guidelines, I have updated a little to have more contrast (make the text readable) but still not a good score.

fixes #429

**This pull request has not been QAed properly yet, it's here just to validate the general direction of this feature. So there might be some bugs lurking.**  If you have extra time, feel free to review it anyway. 🙏


www/src/scss/_theme-moody-blue.scss
```diff
@@ -75,6 +75,11 @@ $highlight-text: #cfd8dc;
     color: $text-base;
   }
 
+  .shortcut-tray { background: darken($bg-main, 10); }
```
inderpreetsingh: will create a mixin for all the repetitive code in all the css files during next css cleanup sprint.

maneetpaul:
@inderpreetsingh - would you like us to merge this into the Alpha? Or would you like to make the changes we discussed over Slack first?

www/js/shortcut_tray.json
```diff
@@ -0,0 +1,26 @@
+{
+    "anandSahib": {
+        "label": "Anand Sahib",
+        "labelType": "roman",
+        "type": "shabad",
+        "ref": 333375
```
tsingh777: i'm guessing this will be updated before the release.

www/js/shortcut_tray.js
```diff
@@ -0,0 +1,49 @@
...
+const { store } = remote.require('./app');
+
+const shortcutTrayContainer = document.querySelector('.shortcut-tray');
+let isShortcutTrayOn = store.getUserPref('slide-layout.display-options.shortcut-tray-on');
+
+const trayItemFactory = (trayItemKey, trayItem) => h(
+  `div.tray-item.${trayItem.labelType}#tray-${trayItemKey}`,
+  {
+    onclick: () => {
+      if (trayItem.type === 'text') {
```
tsingh777: 👍

www/js/shortcut_tray.json
inderpreetsingh: Yeah, this will be changed to`type:bani`, and use banidb to load anand sahib + salok. 🙏

PR #494, opened 2019-02-14
Initializing analytics after app is ready

### LINKED ISSUE #483: Error message upon launch for Mac users

tsingh777:
Fixes #483

### DIFF
--- a/app.js
+++ b/app.js
@@ -45,16 +45,6 @@ const store = new Store({
   defaults: defaultPrefs,
 });
 
-// Retrieve the userid value, and if it's not there, assign it a new uuid.
-let userId = store.get('userId');
-
-if (!userId) {
-  userId = uuid();
-  store.set('userId', userId);
-}
-const analytics = new Analytics(userId, store);
-global.trackEvent = analytics.trackEvent;
-
 const appVersion = app.getVersion();
 
 let mainWindow;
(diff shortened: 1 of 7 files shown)


app.js
```diff
@@ -45,16 +45,6 @@ const store = new Store({
...
 });
 
-// Retrieve the userid value, and if it's not there, assign it a new uuid.
-let userId = store.get('userId');
-
-if (!userId) {
-  userId = uuid();
-  store.set('userId', userId);
-}
-const analytics = new Analytics(userId, store);
```
navdeepsinghkhalsa: What if you left a `let analytics;` declaration here for scope and then assigned it when `app` is ready? It would be good to be consistent with how we access main process variables in the renderer process.

tsingh777: assigning  `analytics` in `app.on('ready',`  does not seem to work when grabbing it with ```analytics } = require('electron').remote.require('./app');``` evaluates to undefined. 

we used ```remote.getGlobal``` in overlay.js and it is in the Docs (https://electronjs.org/docs/api/remote). Maybe we can make that the preferred way?

navdeepsinghkhalsa: Ah, I feared that might be the case. Good to know! I'm glad we tried it and can move to globals in a future enhancement.

Issue #218, opened 2018-07-22
Automated Testing


lbk003:
Epic: Pre-commit should run a set of automated tests with adequate test coverage to ensure quality and backward compatibility as changes are made

DoD
- Identify and hook up a test framework
- Create a test plan and get it reviewed from core dev team.
- Write Unit Tests for existing code
- Add pre-commit hooks to run the test
- Enhance documentation for dev team to use, maintain, and enhance this set of tests

PR #496, opened 2019-02-18
Add Spectron and Jest for integration tests

tsingh777:
RFC For testing framework. Added some basics test with hopes to expand. 
Relates to https://github.com/KhalisFoundation/sttm-desktop/issues/218

Run locally with `npm run test:integration`

Pending Task:
- [x] pass in the built app path so we can run tests against mac build and windows build
- [x] Integrate with CI (https://github.com/electron/spectron#on-travis-ci)
     - this might require building the app first ~without packaging~.
- [ ] Add windows app testing scripts
- [ ] Investigate how to include Testing DB so more scenarios can be tested.
- [ ] come up with meaningful scenarios to write tests for.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,10 @@
     "build-css:main": "node-sass --output-style compressed --include-path www/src/scss www/src/scss/styles.scss www/assets/css/bundle.css",
     "build-css:obs": "node-sass --output-style compressed --include-path www/src/scss/obs www/src/scss/obs/styles.scss www/obs/bundle.css",
     "watch-css": "nodemon --watch www -e scss -x \"npm run build-css\"",
-    "postinstall": "electron-builder install-app-deps"
+    "postinstall": "electron-builder install-app-deps",
+    "test:pr": "npm run test && npm run pack:mac && npm run test:integration:mac",
+    "test:integration:mac": "TESTING_APP_PATH=./builds/mac/SikhiToTheMax.app/Contents/MacOS/SikhiToTheMax jest",
+    "test:integration:debug": "TESTING_APP_PATH=./builds/mac/SikhiToTheMax.app/Contents/MacOS/SikhiToTheMax node $NODE_DEBUG_OPTION node_modules/jest/bin/jest"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -23,7 +23,10 @@
     "build-css:main": "node-sass --output-style compressed --include-path www/src/scss www/src/scss/styles.scss www/assets/css/bundle.css",
     "build-css:obs": "node-sass --output-style compressed --include-path www/src/scss/obs www/src/scss/obs/styles.scss www/obs/bundle.css",
     "watch-css": "nodemon --watch www -e scss -x \"npm run build-css\"",
-    "postinstall": "electron-builder install-app-deps"
+    "postinstall": "electron-builder install-app-deps",
+    "test:pr": "npm run test && npm run pack:mac && npm run test:integration:mac",
```
tsingh777: Adds about 5mins to the build

tests/integration/app.test.js
```diff
@@ -0,0 +1,89 @@
...
+      const mainTabString = await app.client.getAttribute('#history-tab-content', 'class');
+      const mainTab = mainTabString.split(' ');
+      expect(mainTab).toContain('active');
+
+      await app.client.click('#themes-tab');
+      let classListString = await app.client.getAttribute('#themes-tab-content', 'class');
+      let classList = classListString.split(' ');
+      expect(classList).toContain('active');
+
+      await app.client.click('#insert-tab');
```
inderpreetsingh: Do you think these selectors might change when we refactor?

tests/integration/app.test.js
```diff
@@ -0,0 +1,89 @@
...
+  afterEach(() => {
+    if (app && app.isRunning()) {
+      return app.stop();
+    }
+    return false;
+  });
+
+  test('should test true is true', async () => { // true
+    expect(true).toBe(true);
+  });
```
inderpreetsingh: Possibly noob question, What would be the purpose of this?

tests/integration/app.test.js
```diff
@@ -0,0 +1,89 @@
...
+      expect(classList).toContain('active');
+    });
+  });
+
+  describe('Themes', () => {
+    test('app contains default theme', async () => {
+      const classListString = await app.client.getAttribute('body', 'class');
+      const classList = classListString.split(' ');
+
+      expect(classList).toContain('light-theme');
```
inderpreetsingh: should default theme come from a common constant variable?

tests/integration/app.test.js
tsingh777: was there to test that the test were running properly. can remove it.

tests/integration/app.test.js
tsingh777: I can make it read from the `defaults.json`. will make that change

tests/integration/app.test.js
tsingh777: if they change, the test can be updated

tsingh777:
Really out of date now. Will revisit after refactor.

PR #500, opened 2019-02-19
#486 Fixing the delay in changing lines when next line feature is enabled

### LINKED ISSUE #486: Long animation when using next line
There is a long animation when changing lines while the "next line" toggle is enabled. 

This animation doesn't exist when the toggle is turned off. Need to remove the animation so it is consistent. 

nextline_animation

sukhpalsingh:
Fixes #486 

There was animation added when the next line feature is enabled which was causing the delay in gurmukhi, translation and transliteration appearing on the screen.

Removing the animation in the css fixes the delay issue as shown below:

After the fix:

### DIFF
--- a/www/src/scss/_display-next-line.scss
+++ b/www/src/scss/_display-next-line.scss
@@ -40,13 +40,6 @@
         padding-top: 0.5em;
       }
     }
-
-    .gurbani,
-    .translation,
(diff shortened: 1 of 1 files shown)


sukhpalsingh:
Already commited by another user so closing it now.

PR #501, opened 2019-02-19
#487 Making quick tools to overlay on the slide

### LINKED ISSUE #487: Quick tools should overlay on top of the slide, not push everything over
quick tools

sukhpalsingh:
Fixes #487 

Making the quick tools to overlay on the slide rather than pushing the slide.

### DIFF
--- a/www/src/scss/viewer.scss
+++ b/www/src/scss/viewer.scss
@@ -301,7 +301,7 @@ h2 {
 
 .viewer-controls {
   align-items: center;
-  background-color: #ccc;
+  background-color: #e0e0e0;
   display: none;
   flex-direction: column;
   height: 0;
--- a/www/src/scss/viewer.scss
+++ b/www/src/scss/viewer.scss
@@ -297,11 +297,6 @@ h2 {
   &.webview .vc-toggle-icon {
     display: flex;
   }
-
-  .deck.vc-open {
-    right: #{$viewer-padding};
-    width: calc(100% - #{$viewer-padding} * 2 - 28vh);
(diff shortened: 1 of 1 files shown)


inderpreetsingh:
Thank you @sukhpalsingh  veerji. This looks good to me code wise, needs to get design input from @saintsoldierx

maneetpaul:
@sukhpalsingh @inderpreetsingh can we make the overlay slightly less transparent? So we can't see the text as clearly that is behind the quick tools

sukhpalsingh:
Made changes to the background opacity and color so that it does not look too transparent as below:

tarunsingh5:
really like this vs the pushing the side in - @saintsoldierx can you give us a quick thumbs up

maneetpaul:
Merging into alpha for QA so we can include in next release. We can adjust if needed.

saintsoldierx:
It looks great - but I preferred it to be most transparent. Can we change the opacity to make it xparent please. Use case is if it is being used in single screen mode and user needs to change the font size, they can do so without disrupting people who are reading too much.

maneetpaul:
> It looks great - but I prefer it to be more transparent. Can we change the opacity to make it xparent please. Use case is if it is being used in single screen mode and user needs to change the font size, they can do so without disrupting people who are reading too much.

We can make it transparent, but let's do an opacity a bit more solid than what it was before? The text from the slide was blending in too much with the tools text. 

Also, our analytics show that single display mode is rarely (if at all) used. So I don't think this will be an issue.

Issue #507, opened 2019-02-22
Preference are getting lost after closing the app


sukhpalsingh:
It used to load the saved preferences even after reopening the application after close but in latest version it is not loading the saved preferences any more.

The default preference are getting loaded and database is getting download on every start.

The issue has been identified in windows 10 and linux for 4.4 version (production release)

maneetpaul:
Thank you for letting us know @sukhpalsingh. We just released 4.4.1 that fixes this.

PR #508, opened 2019-02-22
Read preferences file without create flag

tsingh777:
Fixes https://github.com/KhalisFoundation/sttm-desktop/issues/507

the flag prop was overriding the prefs json file on load.


maneetpaul:
Need to merge master into dev (4.4.1 released that fixes this)

PR #513, opened 2019-02-27
Surfacing first ceremony (Anand Sahib with Salok)

### LINKED ISSUE #514: Backend work for ceremonies.
Separating the backend work fro #428 that's due release in 5.0

inderpreetsingh:
Anand sahib with salok works now but there are some problems.

* I don't have transliterations in realm DB I was given.
* There are still no ceremonies in sqlite db i have.
* The translations and teeka were also changed in schema, so I updated existing code. I just hope that was intentional (because it will break things with existing db).-

*update:*
fixes #514 #515 

Database structure in realm and sqlite were different so had to add conditionals to conform. 🙏


navdeepsinghkhalsa:
- Transliterations will now be done on-the-fly using `anvaad-js` https://www.npmjs.com/package/anvaad-js
- Please make sure you have the most recent DB. @ManjotS added them last week
- Yes. It was intentional :)

navdeepsinghkhalsa:
GitHub's introduced a cool new feature called Draft Pull Requests. Could you please use those in the future instead of `[WIP]`? :)
https://github.blog/2019-02-14-introducing-draft-pull-requests/

www/js/shortcut_tray.json
```diff
@@ -2,8 +2,8 @@
     "anandSahib": {
         "label": "Anand Sahib",
```
tsingh777: just wonder, should we some how specify this is 6 pauri anand sahib (with salok)?

tsingh777: cc. @maneetpaul

www/js/banidb/realm-search.js
```diff
@@ -161,6 +161,30 @@ const loadShabad = ShabadID => (
...
+ * @returns {object} Returns array of objects for each line
+ * @example
+ *
+ * loadCeremony(3);
+ * // => [{ Gurmukhi: 'jo gurisK guru syvdy sy puMn prwxI ]', ID: 31057 },...]
+ */
+
+const loadCeremony = ceremonyID => (
+  new Promise((resolve, reject) => {
+    Realm.open(realmConfig)
```
navdeepsinghkhalsa: Could you please do the check to see if it's been initialized first?

inderpreetsingh: Sure. Sorry, missed it 🙈

www/js/shortcut_tray.json
maneetpaul: @inderpreetsingh can we label it "Anand Sahib + Salok"

www/js/controller.js
```diff
@@ -477,7 +478,20 @@ module.exports = {
     global.platform.ipc.send('clear-apv');
   },
 
-  sendLine(shabadID, lineID, Line, rows) {
+  remapLine(rawLine) {
+    const Line = Object.assign(rawLine, {});
```
navdeepsinghkhalsa: `Object.assign({}, rawLine);`

inderpreetsingh: I am not sure why but that didn't work for me. 🙏 🤔 

Looking into it again.

PR #516, opened 2019-02-28
CustomBackground: Checks MimeType of the image and not just extension

inderpreetsingh:
I promised this during custom background image PR, even sent a PR but had to close it due to other priorities. I had some time today, so sending a PR again to check mime type of the image and not just the extension.


www/js/theme_editor.js
```diff
@@ -37,10 +39,21 @@ const uploadErrorNotification = (message, timeout = 3000) => {
...
 
 const imageCheck = (filePath) => {
-  const acceptedExtensions = ['.png', '.PNG', '.jpg', '.JPG', '.jpeg', '.JPEG'];
-  const extension = path.extname(filePath);
+  const acceptedExtensions = ['png', 'jpg'];
+  const acceptedMimeTypes = ['image/png', 'image/jpeg'];
 
-  return acceptedExtensions.includes(extension);
+  try {
+    const buffer = readChunk.sync(filePath, 0, 12);
```
tsingh777: there's a bug on line 179 that causes this to fail
```
 sortedFiles.forEach((file) => {
        const fullPath = path.resolve(userBackgroundsPath, file);
        if (imageCheck(file)) {
          recentBgsContainer.appendChild(recentSwatchFactory(fullPath));
        } else {
          fs.unlink(fullPath, (deleteError) => {
            if (deleteError) uploadErrorNotification(`Unable to delete a file. - ${deleteError}`);
          });
        }
      });
```
`imageCheck(file)` should be `imageCheck(fullPath)`.

tarunsingh5:
@inderpreetsingh

www/js/theme_editor.js
inderpreetsingh: Aah, I see. Thanks 🙏

PR #517, opened 2019-03-01
Realm evergreen

navdeepsinghkhalsa:
The schema definition is now packaged with the database and the `schemaVersion` will be updated in that definition to prevent conflicts


www/js/desktop_scripts.js
```diff
@@ -133,9 +136,12 @@ module.exports = {
                     fs.unlinkSync(dbCompressed);
                     // Replace current DB file with new version
                     fs.renameSync(newDBPath, dbPath);
+                    if (dbPlatform === 'realm') {
+                      fs.renameSync(newDBSchema, dbSchema);
+                    }
                     module.exports.initDB();
                     // Delete old DBs
-                    const oldDBs = ['data.db'];
+                    const oldDBs = ['data.db', 'sttmdesktop.realm', 'sttmdesktop.realm.lock'];
```
tsingh777: should `sttmdesktop.realm.management` folder also be cleaned up?

www/js/banidb/realm-search.js
```diff
@@ -1,7 +1,29 @@
...
+const electron = require('electron');
+const path = require('path');
 const Realm = require('realm');
-const realmDB = require('./realm-db');
+
 const CONSTS = require('./constants');
 
+const { remote } = electron;
+const userDataPath = remote.app.getPath('userData');
+const realmPath = path.resolve(userDataPath, 'sttmdesktop-evergreen.realm');
```
tsingh777: just wondering, is the new one intentionally sparse at the moment? not seeing any translations after checking out this branch.

www/js/banidb/realm-search.js
```diff
@@ -1,7 +1,29 @@
...
+const realmSchemaPath = path.resolve(userDataPath, 'realm-schema-evergreen.json');
+
+const realmSchemaDef = {
```
tsingh777: nit picking- is this the appropriate name? this object is more than just the schemaDef, its all the config you can pass to the `open` function.
https://realm.io/docs/javascript/latest/api/Realm.html#.open

www/js/banidb/realm-search.js
```diff
@@ -214,7 +251,7 @@ const getShabad = VerseID => (
  */
 const randomShabad = (SourceID = 'G') => (
   new Promise((resolve) => {
-    Realm.open(realmDB.realmVerseSchema)
+    Realm.open(realmSchemaDef)
```
tsingh777: will opening the DB multiple times have any memory issues? doing some profiling looks like each time a search occurred the number of Realm objects increased (after a few searches by 1mb). Something to look out for incase we get feeback that the app gets progressively slower the more you use it.

www/js/desktop_scripts.js
navdeepsinghkhalsa: We can add it as a TODO. The current implementation doesn't remove directories.

www/js/banidb/realm-search.js
navdeepsinghkhalsa: It is intentionally sparse, yes. Unicode and Transliterations were removed and will be done on-the-fly if needed with `anvaad-js`. Translations should be there, but all combined as a JSON object in the `Translations` column

www/js/banidb/realm-search.js
navdeepsinghkhalsa: We can make it `realmSchemaConfig`

www/js/banidb/realm-search.js
navdeepsinghkhalsa: I'm glad you looked into this. Let's see what the best practice is.

www/js/banidb/realm-search.js
tsingh777: I see. Looks like we need to update some code use display it properly.

PR #518, opened 2019-03-06
Fixes `sqlite` db object not having a `dbSchema`

inderpreetsingh:
Sqlite db object had no key called `dbSchema` but independent of dbPlatform dbSchema was being used to resolve path.


www/js/desktop_scripts.js
```diff
@@ -34,11 +35,14 @@ if (platform === 'win32') {
   }
 }
 
+const dbSchemaPath = schemaPath =>
+  !database[dbPlatform].dbSchema || path.resolve(schemaPath, database[dbPlatform].dbSchema);
+
 const dbPath = path.resolve(userDataPath, database[dbPlatform].dbName);
-const dbSchema = path.resolve(userDataPath, database[dbPlatform].dbSchema);
+const dbSchema = dbSchemaPath(userDataPath);
```
tsingh777: that really interesting, this would evaluate to `true` for sqlite. you then pass that to `fs.existsSync()` which only takes `path <string> | <Buffer> | <URL>` [DOCS](https://nodejs.org/docs/latest-v8.x/api/fs.html#fs_fs_existssync_path). I'm wondering if that would have any issues down the road. its returning false now, so Its fine for now.

www/js/search.js
```diff
@@ -690,7 +690,7 @@ module.exports = {
 
   printShabad(rows, ShabadID, LineID) {
     const lineID = LineID || rows[0].ID;
-    const shabadID = ShabadID || rows[0].Shabads ? rows[0].Shabads[0].ShabadID : '';
+    const shabadID = ShabadID || (rows[0].Shabads ? rows[0].Shabads[0].ShabadID : '');
```
tsingh777: 👍

PR #521, opened 2019-03-08
Port release script to node and make cross-platform

navdeepsinghkhalsa:


packaging/release
```diff
@@ -0,0 +1,50 @@
...
+const exec = util.promisify(require('child_process').exec);
+const { argv: { branch, platform, token } } = require('yargs');
+const checkVersion = require('./check-version');
+const updateRemoteVersion = require('./update-remote');
+
+const release = async () => {
+  const track = branch === 'dev' ? 'alpha' : 'beta';
+
+  // Get all tags, sort in desc order, and get latest from current track
+  const gitTags = await git().tags();
```
tsingh777: you could pass a sort function to the tags method.

packaging/release
```diff
@@ -0,0 +1,50 @@
...
+  const gitTag = gitTags.all.reverse().find(tag => tag.indexOf(track) > -1);
+  console.log(`Last Git tag: ${gitTag}`);
+
+  // Get a new release version and update package.json
+  const releaseVersion = checkVersion(branch, gitTag);
+  console.log(`New release version: ${releaseVersion}`);
+
+  console.log('TODO: stream output for the next command');
```
tsingh777: can you elaborate a bit?

packaging/release
navdeepsinghkhalsa: When I do that, I get an old tag that starts with 'v' that always comes up first.

packaging/release
navdeepsinghkhalsa: Line 26 is only going to log to the console once `npm run dist:platform` is done in the current way, so anyone watching will be waiting a while. In the future, we should use `spawn` instead of `exec` from `child_process`

navdeepsinghkhalsa: That allows you to pipe the output from the command to `console.log`

packaging/check-version.js
```diff
@@ -7,45 +7,44 @@ const packageJSON = require('../package.json');
...
-  fs.writeFileSync(path.resolve(__dirname, '..', 'package.json'), JSON.stringify(packageJSON));
-}
-
-console.log(currentRelease);
+  if (branch === 'dev' || branch === 'master') {
+    const track = branch === 'dev' ? 'alpha' : 'beta';
+    // If the release version is the same and the last one was either an alpha or beta,
+    // increment the prerelease version
+    if (
+      semver.eq(lastReleaseMain, currentReleaseMain)
```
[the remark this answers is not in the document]
navdeepsinghkhalsa: We log the version on line 21 of the `release` file

Issue #437, opened 2019-01-11
Update Settings to include Bani Length for SG


maneetpaul:
When we surface SG there needs to be a way to choose Bani Length

maneetpaul:
@inderpreetsingh  - can we change this so it is a dropdown with options for "Short", Medium", "Long", and "Very Long"

Need to stay consistent with the Sundar Gutka app

inderpreetsingh:
@maneetpaul got it.

Issue #433, opened 2019-01-11
Add Anayltics for each Bani in SG


maneetpaul:
Need to be able to track the number of people who click a Bani from the SG menu vs the number of people who search for it manually

PR #532, opened 2019-03-12
Sunder gutka toolbar

### LINKED ISSUE #354: Surface Sundar Gutka
Surface all Banis avaialble in Sundar Gutka

inderpreetsingh:
fixes #354  #433 #437 
* [x] Add sunder gutka UI presenter mode
* [x] Add sunder gutka UI non-presenter mode
* [x] Add realm support
* [x] Add sqlite support
* [x] Adds close button
* [x] Clean/optimize code
* [x] talk and figure out about special verses / lines with different db structure.


www/js/toolbar.js
```diff
@@ -0,0 +1,109 @@
...
+const baniGroupFactory = (baniGroupName) => {
+  const baniGroupClass = baniGroupName.replace(/ /g, '-');
+  return h(
+    'div.bani-group-container',
+    h(`header.bani-group-heading.${baniGroupClass}-heading`, baniGroupName),
+    h(`div.bani-group.${baniGroupClass}`),
+  );
+};
+
+const extrasTile = (tileType, row) => h(
```
tsingh777: nit picking: should this be `tileFactory` or something?

www/js/search.js
```diff
@@ -663,6 +663,24 @@ module.exports = {
...
 
+  loadBani(BaniID) {
+    const $shabadList = this.$shabad || document.getElementById('shabad');
+    $shabadList.innerHTML = '';
+    banidb.loadBani(BaniID)
+      .then(rowsDb => {
+        const rows = rowsDb.map((rowDb) => {
+          let row = rowDb;
+          // when object from db is not a verse itself
+          if (rowDb.Verse) row = rowDb.Verse;
```
tsingh777: nit picking: can we be consistent with if statements and do one liners with curly brackets as well?

www/js/toolbar.js
```diff
@@ -0,0 +1,109 @@
...
+  init() {
+    toolbarItems.forEach((toolbarItem) => {
+      $toolbar.appendChild(toolbarItemFactory(toolbarItem));
+      $toolbar.appendChild(closeOverlayUI);
+      $baniExtras.appendChild(baniGroupFactory('nitnem banis'));
+      $baniExtras.appendChild(baniGroupFactory('popular banis'));
+      banidb.loadBanis().then(rows => printBanis(rows));
+    });
+  },
+};
```
tsingh777: `toolbar.json` seems to be unused. remove?

tsingh777:
Looks really good! 
Some observations:
 - long banis have severely degraded the performance of the spacebar shortcut. it locks up the UI. We have to fix that before releasing.
- when in non presenter view and the window is collapsed, it looks like:  image probably needs some padding or a smaller icon?
- some lines just end up being blank: 
image 
- nitnem banis probably don't need to be in popular banis (arti arta)
image
- at 1040px-1050px width it looks like this 
image
- The what's new modal is now showing up behind the other windows:
image
- should sundar gutka populate history? It isn't with these changes c c.@maneetpaul 
- is there a duplicate line in the sundar gutka anand sahib? 
image
- is `Anand Sabhib & Salok` a good enough differentiator for 6 Pauris vs 40 Pauris? cc. @maneetpaul 


---------
- Feature request: can we have translit somewhere in the bani list? 
- Feature request: with long banis, it going to become difficult scrolling to find where in the bani the user needs to present. should we add a search or something like the bookmark in sundar gutka that takes us a specific place in the bani.

maneetpaul:
For the Nitmem section, please have: 

-- Japjee Sahib
-- Jaap Sahib
-- Tav Prasaad Svwaiye
-- Benti Chupai Sahib
-- Anand Sahib
-- Rehras Sahib
-- Sohila Sahib

For popular, please use:

-- Asa Kee Vaar
-- Aarti
-- Salok M. 9
-- Sukhmani Sahib


@tsingh777 responses to some of your notes: 

- For now, leave sundar gutka out of history. We will revisit that in a future Sprint along with a few other fixes/updates 
- Yes, I believe we differentiate the two Anand Sahibs well with the current wording 
- If it is an easy add, yes, I agree we should add a toggle to the bani list to show it in english vs gurmukhi. If this will take a lot more time, we will save it for future Sprint. @inderpreetsingh please let us know. 
- I think people will use normal search when they are looking for a specific line inside of a ban (rather than open the entire bani and scroll for it). We should see what kind of feedback we get. Will keep an eye on this. Additionally, people can type the first letter of the line they are looking for once the bani is open and skip through pretty quickly like that (this was an old STTM2 feature). 

Thank you for the comments!

inderpreetsingh:
Fixed everything except DB related bugs or new features. 🙏 

 Using translit I get these spellings, do we want to go ahead with these spellings?
and we don't have english names of banis in database. We have tokens but they don't have spaces in bani names.

Screenshot 2019-03-19 at 2 16 15 PM

inderpreetsingh:
I say. 👍 +1 to adding bookmarks. I think we have them in database too. 🙏

maneetpaul:
I am not a fan of the current English list. Some of the spellings look really strange and we need to make each bani fit on one line so it looks cleaner. Let's release 5.0 with Gurmukhi only for now, and we'll add the English names for each Bani to the DB properly. We'll add a toggle in a future update to switch between the two. 

Regarding pagination, I am against requiring the user needing to do anything extra in order to go through the entire bani (they should be able to go through the whole thing by pressing the down arrow).  Let's try to isolate the issue and see if it is specific to a certain bani.

www/index.html
```diff
@@ -53,9 +53,17 @@
...
   </div>
   <div id="app-frame">
-    <webview id="main-viewer" src="viewer.html" nodeintegration></webview>
+    <webview id="main-viewer" src="viewer.html" nodeintegration class="base-ui"></webview>
     <div id="navigator" class="noselect">
-      <div id="main-ui">
+      <div class="focus-overlay hidden overlay-ui common-overlay"> </div>
+      <div id="toolbar">
+        <div class="bani-list hidden overlay-ui sg-overlay"> 
+          <header class="navigator-header"> <span> Sunder Gutka </span> </header>
```
tsingh777: `Sundar Gutka`

www/js/toolbar.js
```diff
@@ -0,0 +1,130 @@
...
+  );
+};
+
+const extrasTileFactory = (tileType, row) => h(
+  `div.extras-tile.${tileType}`,
+  {
+    onclick: () => {
+      global.core.search.loadBani(row.ID);
+      toggleOverlayUI();
+      analytics.trackEvent('sunderGutkaBanis', tileType, row.token);
```
tsingh777: should be `row.Token`

www/js/toolbar.js
```diff
@@ -0,0 +1,130 @@
...
+    }
+    if (popularBanis.includes(row.ID)) {
+      baniTag = 'popular';
+      $popularBanis.appendChild(extrasTileFactory('popular-bani', row));
+    }
+    const $bani = h(
+      'li.sunder-gutka-bani',
+      {
+        onclick: () => {
+          analytics.trackEvent('sunderGutkaBanis', row.token);
```
tsingh777: `row.Token`

tsingh777:
Sukhmani Sahib is the least performant. 

The following function from `index.js` needs to be optimized. I will look into it. 
```
function spaceBar(e) {
  const mainLineID = search.$shabad.querySelector('a.panktee.main').dataset
    .lineId;
  const currentLineId = search.$shabad.querySelector('a.panktee.current')
    .dataset.lineId;

  let newLineId = mainLineID;

  if (mainLineID === currentLineId) {
    let done = false;
    search.$shabad.querySelectorAll('a.panktee').forEach((item) => { //2099 lines for Sukhmani sahib
      if (!item.classList.contains(['seen_check']) && !done) {
        newLineId = item.dataset.lineId;
        done = true;
      }
    });
  }

  highlightLine(newLineId);
  e.preventDefault();
}
```

maybe we can look into adding virtual scrolling as well. not sure if a libaray exists that can help here. cc. @navdeepsinghkhalsa @bogas04

tsingh777:
@inderpreetsingh https://github.com/inderpreetsingh/sttm-desktop/pull/2

www/js/defaults.json
```diff
@@ -4,7 +4,7 @@
       "theme": "light-theme",
       "themebg": "",
       "layout": {
-        "presenter-view": true
+        "presenterview": true
```
tsingh777: why did this change?

www/js/viewer.js
```diff
@@ -215,6 +215,8 @@ const createCards = (rows, LineID) => {
   const cards = [];
   const lines = [];
   const shabad = {};
+  // empty the object from previous values
+  // Object.keys(shabad).forEach(function(key) { delete shabad[key]; });
```
tsingh777: remove?

www/js/toolbar.js
```diff
@@ -0,0 +1,130 @@
+const h = require('hyperscript');
+const banidb = require('./banidb');
+const { remote } = require('electron');
+const anvaad = require('anvaad-js');
+
+const analytics = remote.getGlobal('analytics');
+
+const toolbarItems = ['sunder-gutka'];
+const nitnemBanis = [2, 4, 6, 9, 10, 20, 21, 23];
+const popularBanis = [90, 30, 31, 22];
```
tsingh777: there's a `bani_groups.json` that is empty. remove?

www/js/defaults.json
```diff
@@ -35,6 +35,9 @@
       "autoplay-options": {
         "autoplayTimer": 10
       },
+      "sunder-gutka": {
+        "bani-length": 4
```
tsingh777: let's use the same default as the sundar gutka app? I think that one is 2 or 3. cc @maneetpaul @tarunsingh5

tsingh777: can be followed up in separate pr

maneetpaul: The new SG update requires users to choose a length upon the first app launch. Prior to that, the default was "Very Long".

tsingh777: maybe we can implement something similar and make it a choose on first install

maneetpaul: Let's see what the UX is like during Alpha testing and we'll go from there. Bani length only impacts ਬੇਨਤੀ ਚੌਪਈ ਸਾਹਿਬ, ਰਹਰਾਸਿ ਸਾਹਿਬ, ਆਰਤੀ, and ਸੋਹਿਲਾ ਸਾਹਿਬ

www/js/index.js
```diff
@@ -20,8 +21,9 @@ function hideSlide() {
   global.controller.sendText('');
 }
 
-function highlightLine(newLine) {
-  const $line = search.$shabad.querySelector(`#line${newLine}`);
+function highlightLine(newLine, nextLineCount = null) {
```
tsingh777: nit: the variables names (`nextLineCount`) here are bit confusing.

PR #534, opened 2019-03-12
replaces punjabiUni with punjabi

inderpreetsingh:
Unicode dint work on chromecast, so I replaced it with non unicode version. 🙏

### DIFF
--- a/www/obs/index.html
+++ b/www/obs/index.html
@@ -11,7 +11,7 @@
     var previewData = {
       Gurmukhi: 'ik®pw kry jy AwpxI qw hir rKw aur Dwir ]1]',
       English: 'If the Lord shows His Mercy, then I keep Him enshrined within my heart. ||1||',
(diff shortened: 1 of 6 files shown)


www/obs/index.html
```diff
@@ -11,7 +11,7 @@
     var previewData = {
       Gurmukhi: 'ik®pw kry jy AwpxI qw hir rKw aur Dwir ]1]',
       English: 'If the Lord shows His Mercy, then I keep Him enshrined within my heart. ||1||',
-      PunjabiUni: 'ਜੇ ਪਰਮਾਤਮਾ ਆਪਣੀ ਮੇਹਰ ਕਰੇ, ਤਾਂ ਮੈਂ ਭੀ ਉਸ ਦਾ ਨਾਮ ਹਿਰਦੇ ਵਿਚ ਪਰੋ ਰੱਖਾਂ ।੧।',
+      Punjabi: 'ਜੇ ਪਰਮਾਤਮਾ ਆਪਣੀ ਮੇਹਰ ਕਰੇ, ਤਾਂ ਮੈਂ ਭੀ ਉਸ ਦਾ ਨਾਮ ਹਿਰਦੇ ਵਿਚ ਪਰੋ ਰੱਖਾਂ ।੧।',
```
navdeepsinghkhalsa: This should be in Gurmukhi

navdeepsinghkhalsa:
A similar update will have to be made to the Chromecast app to support *both* unicode and ASCII, as pre-v4.5 versions of the app will be sending unicode

Issue #542, opened 2019-03-17
Only allow one instance of STTM to be running


maneetpaul:
Users reporting that you can open multiple instances of STTM which is causing issues at Gurdwaras

PR #543, opened 2019-03-19
Only allow one instance of STTM

tsingh777:
#542 - Follows the example from https://github.com/electron/electron/blob/v0.36.10/docs/api/app.md#appmakesingleinstancecallback

### DIFF
--- a/app.js
+++ b/app.js
@@ -266,7 +266,9 @@ const emptyOverlay = () => {
 const shouldQuit = app.makeSingleInstance(() => {
   // Someone tried to run a second instance, we should focus our window.
   if (mainWindow) {
-    if (mainWindow.isMinimized()) mainWindow.restore();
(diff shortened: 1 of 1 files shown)


app.js
```diff
@@ -263,6 +263,18 @@ const emptyOverlay = () => {
...
+const shouldQuit = app.makeSingleInstance(() => {
+  // Someone tried to run a second instance, we should focus our window.
+  if (mainWindow) {
+    if (mainWindow.isMinimized()) mainWindow.restore();
+    mainWindow.focus();
+  }
+});
+
+if (shouldQuit) {
+  app.quit();
```
navdeepsinghkhalsa: The example includes a `return;` after this line. Is it needed?

tsingh777: It is not.

PR #547, opened 2019-03-21
Sunder gutka with themes

inderpreetsingh:
Adds support for themes in sunder gutka interface.


.gitignore
```diff
@@ -10,4 +10,4 @@ data.db
 npm-debug.log*
 .idea/
 .DS_Store/
-
+www/.DS_Store
```
tsingh777: `.DS_Store` should cover everywhere. no?

inderpreetsingh: I thought so too. Apparently, it did not. 

I looked up and there's **/.DS_Store syntax, but need to check.

PR #548, opened 2019-03-21
Resurface vishraams

### LINKED ISSUE #436: Surface Vishraams (PENDING DB WORK)
Ticket to track surfacing of vishraams once DB is updated

### LINKED ISSUE #549: Allow different vishraam sources
sttm living vishraams, igurbani, sttm 2

inderpreetsingh:
Screenshot 2019-03-22 at 2 45 27 AM

Bringing vishraams back. 🎉 🎈 

Fixes #436 

Fixes #549 

Just need to fix the DB and then i will make this PR merge-able.


maneetpaul:
Can we include work from #549 in this PR as well? 

Would like to have Vishraam options and Bani Length options under the same heading in the Settings. Is it possible to consolidate the Vishraam options to a drop-down instead of taking up two lines with a checkbox?

inderpreetsingh:
@maneetpaul yes we can add vishraam sources.

Regarding having dropdown settings, we can do it, it's a tiny bit tougher than using current supported settings format (which does not include a dropdown). I can definitely do it, just wanted to let you know that it's not as easy as adding currently supported formats.

maneetpaul:
@inderpreetsingh the dropdown PR has been merged so we can continue on this one! 

To recap: 

- Vishram style options should be a dropdown
- Vishram source also added as an option with a dropdown (please let me know if this is going to require more than 1pt of work. We can move it to future Sprint if needed)
- Vishram settings to live under the Gurbani Settings heading (right underneath Bani length)

inderpreetsingh:
Found another bug, need to factor in change of number of verses when we change the bani length and reload bani accordingly.

inderpreetsingh:
Fixes #549 as well.

www/js/settings.json
```diff
@@ -92,8 +92,29 @@
...
+              "colored-words": "Colored Words",
+              "gradient-bg": "Gradient Background"
+            }
+          },
+          "vishraam-source": {
+            "title": "Vishraam Source",
+            "options": {
+              "sttm2": "STTM 2",
+              "igurbani": "iGurbani",
+              "sttm": "STTM Living"
```
tsingh777: Are we going to explain what these sources mean (where they come from)? some context in the help would be useful @maneetpaul

www/js/settings.json
```diff
@@ -92,8 +92,29 @@
...
+            "options": {
+              "colored-words": "Colored Words",
+              "gradient-bg": "Gradient Background"
+            }
+          },
+          "vishraam-source": {
+            "title": "Vishraam Source",
+            "options": {
+              "sttm2": "STTM 2",
+              "igurbani": "iGurbani",
```
tsingh777: need to give some credit here (also in the help)?

www/js/settings.json
tsingh777: also that they only show for certain themes.

www/js/viewer.js
```diff
@@ -215,24 +214,42 @@ const createCards = (rows, LineID) => {
   const cards = [];
   const lines = [];
   const shabad = {};
-  // empty the object from previous values
+
   Object.keys(rows).forEach((key) => {
     row = rows[key];
```
tsingh777: `row` should be declared here. add a `const`

www/src/scss/viewer.scss
```diff
@@ -1,7 +1,42 @@
 $viewer-gurmukhi-font-family: 'gurbaniakharthick';
 $viewer-padding: 30px;
-$bisram-main: #c0392b;
-$bisram-yamki: #ffc500;
+
+@mixin vishraamStyles() {
```
tsingh777: 👍

www/js/viewer.js
```diff
@@ -215,24 +214,42 @@ const createCards = (rows, LineID) => {
...
     lines.push(row.ID);
-    // const gurmukhiShabads = row.GurmukhiBisram.split(' ');
     const gurmukhiShabads = row.Gurmukhi.split(' ');
+    if (row.Visraam) {
+      const visraams = JSON.parse(row.Visraam);
```
tsingh777: should this should be in a `try`? just in case a downloaded db pushes something wrong or the contract changes, we wouldn't want the app to break immediately. cc @navdeepsinghkhalsa

www/js/settings.json
maneetpaul: Yes! I will update the help to include the details for this

www/js/settings.json
maneetpaul: I'm assuming you mean credit to the other orgs for the other vishraam sources. Yes I will update this info in the help.

www/js/viewer.js
inderpreetsingh: Oh yes, it should be. Thanks. 🙏

PR #553, opened 2019-03-23
adds dropdown as a settings format

### LINKED ISSUE #437: Update Settings to include Bani Length for SG

inderpreetsingh:
Fixes #437 , adds new dropdown format to settings.

### DIFF
--- a/www/js/settings.json
+++ b/www/js/settings.json
@@ -87,7 +87,7 @@
           "bani-length": {
             "title": "Bani Length",
             "options": {
-              "small": "Small",
+              "short": "Short",
               "medium": "Medium",
               "long": "Long",
               "extralong": "Extra Long"
--- a/www/js/settings.json
+++ b/www/js/settings.json
@@ -80,8 +80,8 @@
   "toolbar": {
     "title": "Toolbar",
     "settings": {
-      "sunder-gutka": {
-        "title": "Sunder Gutka Settings",
+      "gurbani": {
+        "title": "Gurbani Settings",
         "type": "dropdown",
         "options": {
           "bani-length": {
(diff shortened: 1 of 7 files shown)


www/js/settings.json
```diff
@@ -74,16 +74,24 @@
...
+      }
+    }
+  },
+  "toolbar": {
+    "title": "Toolbar",
+    "settings": {
       "sunder-gutka": {
-        "title": "Sundar Gutka",
-        "type": "range",
+        "title": "Sunder Gutka Settings",
```
maneetpaul: This should be “Gurbani Settings” since vishraams will be in the same section

www/js/settings.json
```diff
@@ -74,16 +74,24 @@
...
         "options": {
           "bani-length": {
             "title": "Bani Length",
-            "max": 4,
-            "min": 1,
-            "step": 1
+            "options": {
+              "small": "Small",
+              "medium": "Medium",
+              "large": "Large",
```
maneetpaul: This should be Long and then Very Long to stay consistent with Sundar Gutka mobile app

inderpreetsingh: done!

www/js/settings.json
```diff
@@ -74,16 +74,24 @@
...
           }
         }
-      },
-      "sunder-gutka": {
-        "title": "Sundar Gutka",
-        "type": "range",
+      }
+    }
+  },
+  "toolbar": {
```
tsingh777: would it be better if this was closer to fields to display/display options? might not matter.

PR #555, opened 2019-03-27
optimises the performance of loading bani

inderpreetsingh:
* Added virtual scrolling to the controller
* appends slides when needed
* divides `printShabad` into separate modules
* cleans viewer.js a little bit

Now, bani loads much faster, and I see no performance issues. 

`known bugs`
* [x] if you drag the scrollbar on controller super fast to the end, it shows empty div. 
* [x] Possible edge case with bani length (what if next 10 lines are not in currently selected bani length)
* [x] sqlite does not work properly

_update_ : _solved_

`possible enhancements`
* Should I also remove the slides that are way up? Say, I am on verse 100th, should I remove top 30 verses? Or am I going into micro-optimizations?
* I can move the modes to separate constants file, felt like a weird pattern so still thinking on it. Any suggestions are welcome.


inderpreetsingh:
A bug surfaced in this after adding sunder gutka bani length options, so I had to rewrite how bani length is handled. Now instead of hiding/showing verses from CSS, it reloads the whole bani with new bani length attribute.

www/js/search.js
```diff
@@ -663,20 +663,21 @@ module.exports = {
...
-    const baniLengthList = ['short', 'medium', 'long', 'extralong'];
-    const baniLength = store.getUserPref('slide-layout.sunder-gutka.bani-length');
-    const line = rows.find((row) => row.baniLength[baniLengthList[baniLength]]);
-    return line ? line.ID : null;
-  },
-
   loadBani(BaniID) {
     const $shabadList = this.$shabad || document.getElementById('shabad');
+    const baniLength = store.get('userPrefs.toolbar.gurbani.bani-length');
+    const baniLengthCols = {
```
tsingh777: can you comment on this to provide some explanation

www/js/search.js
```diff
@@ -726,76 +719,128 @@ module.exports = {
...
+
+    if (start) {
+      throughput = 10;
+    } else if (lineIndex > throughput) {
+      throughput = lineIndex + 1;
+    }
+
+    const end = start + throughput;
+
+    const lineHeight = 27;
```
tsingh777: Can numbers like these be defined as constants somewhere? having them inline could make refactoring a pain.

www/js/search.js
```diff
@@ -726,76 +719,128 @@ module.exports = {
...
-    if (!apv) {
+    if (!apv && !start) {
       while (shabad.firstChild) {
         shabad.removeChild(shabad.firstChild);
       }
     }
 
-    let lineCount = 0;
-    rows.forEach((item) => {
+    shabad.parentNode.onscroll = (e) => {
```
tsingh777: please add some comments

www/js/search.js
inderpreetsingh: Yes, I agree. I would also like to break down this file into multiple files. I think we will address all these issues in technical debt sprint that's coming up.

Issue #567, opened 2019-04-10
Changelog should open for the first 5 sessions after an update happens


maneetpaul:
This will allow more users to see new updates and be educated on new features

tarunsingh5:
can we include an option to not show again until next update? this may get annoying otherwise, esp if we are doing some small updates

PR #572, opened 2019-04-11
#567 Show changelog multiple times

tsingh777:
https://github.com/KhalisFoundation/sttm-desktop/issues/567
image

### DIFF
--- a/app.js
+++ b/app.js
@@ -320,7 +320,9 @@ app.on('ready', () => {
     // Show changelog if last version wasn't seen
     const lastSeen = store.get('changelog-seen');
     const lastSeenCount = store.get('changelog-seen-count');
-    if (lastSeen !== appVersion || lastSeenCount < maxChangeLogSeenCount) {
+    const limitChangeLog = store.get('userPrefs.app.analytics.limit-changelog');
+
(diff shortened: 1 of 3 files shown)


tsingh777:
@maneetpaul let me know what words/message to put in the option to not show changelog

maneetpaul:
The wording you have in the screenshot looks great

app.js
```diff
@@ -314,7 +319,10 @@ app.on('ready', () => {
     }
     // Show changelog if last version wasn't seen
     const lastSeen = store.get('changelog-seen');
-    if (lastSeen !== appVersion) {
+    const lastSeenCount = store.get('changelog-seen-count');
+    const limitChangeLog = store.get('userPrefs.app.analytics.limit-changelog');
+
+    if (lastSeen !== appVersion || (lastSeenCount < maxChangeLogSeenCount && !limitChangeLog)) {
       openSecondaryWindow('changelogWindow');
```
inderpreetsingh: Can we please add analytics here with timestamp, time difference between when the window opens and when it's closed. That way we can make a guess to see if the changelog was read or not. Like if its too long, that means user just kept it open, and if its too little that means user just immediately closed it. 

As it's already merged, maybe we can do it in 5.1?

PR #576, opened 2019-04-12
Auto format with Prettier

navdeepsinghkhalsa:


package.json
```diff
@@ -142,8 +150,10 @@
       "icon": "assets"
     }
   },
-  "prettier": {
-    "singleQuote": true,
-    "trailingComma": "all"
+  "husky": {
+    "hooks": {
+      "pre-commit": "pretty-quick --staged",
```
inderpreetsingh: not necessary but should we use `lint-staged` to lint only those files that are staged? https://github.com/okonet/lint-staged

navdeepsinghkhalsa: Yes. Good idea

PR #615, opened 2019-05-13
Extends support for vishraam to all themes

### LINKED ISSUE #610: Expand supported themes for Vishraams
See google doc shared on Slack

inderpreetsingh:
Fixes #610

### DIFF
--- a/www/js/settings.json
+++ b/www/js/settings.json
@@ -105,10 +105,7 @@
       "gurbani-options": {
         "type": "switch",
         "options": {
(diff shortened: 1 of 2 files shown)


tsingh777:
Do we need to do anything for the custom backgrounds?

maneetpaul:
Custom backgrounds are following the light theme. For now, those colors should work just fine.

Our metrics show that custom BG's are not being used much. So, for now, we will leave them as a lower priority.

PR #621, opened 2019-05-18
flatten ceremonies array to include verseRange

inderpreetsingh:
Fixes the empty lines that were showing up. Now it shows the required shabads.


inderpreetsingh:
Oh! It does not work for Sqlite yet. _working on it_

www/js/search.js
```diff
@@ -647,25 +647,41 @@ module.exports = {
...
     const $shabadList = this.$shabad || document.getElementById('shabad');
     $shabadList.innerHTML = '';
-    banidb.loadCeremony(ceremonyID).then(rowsDb => {
-      const rows = rowsDb.map(rowDb => {
-        let row = rowDb;
+    $shabadList.dataset.bani = '';
+    try {
+      const rowsDb = await banidb.loadCeremony(ceremonyID);
+      const rows = await Promise.all(
+        rowsDb.map(async rowDb => {
```
tsingh777: there's no `await` used  in this this call back. doesn't need to be `async`.

inderpreetsingh: Thanks, fixed it.

PR #630, opened 2019-05-26
Update sync code to use request-promise and send host

tsingh777:
Sets up code to get a connection. exposing to app is still pending.


Gauravjeetsingh:
@tsingh777 
For having a sundar gutka sync, can we add the following things in data sent to sttm-web:
- type: which can have values like baani, shabad, ceremony
- id: it can be a shabad id, bani id, or a ceremony id
- length: length of baani, since we have short, medium, long and extra long baani lengths on sttm-desktop.

We can treat `id` as shabad id if it's type is Shabad, and bani id if it's a bani.

tsingh777:
@Gauravjeetsingh @inderpreetsingh 

Added the extra props for bani's. For loading the bani, currently desktop will send the ids such as 1, 2, etc. When a line is sent while going through(need to update) the bani, the id we have is like "japji-short", "anand-short" etc. Do we need to agree one format or can web take either or and translate?

www/js/menu.js
```diff
@@ -314,6 +317,61 @@ const announcementSlideButton = h(
...
+        const onlineVal = await isOnline();
+        if (onlineVal) {
+          const code = await tryConnection();
+          if (code) {
+            document.getElementById('remote-sync-icon').style.color = 'green';
+            createSyncModal(`<h2>Sync code: ${code}</h2>`, code);
+          }
+        } else {
+          createSyncModal(`<h2>Must be online to use this feature.</h2>`, '');
+        }
```
bogas04: Add a try catch just in case.

www/js/menu.js
```diff
@@ -314,6 +317,61 @@ const announcementSlideButton = h(
...
+      onclick: async () => {
+        const onlineVal = await isOnline();
+        if (onlineVal) {
+          const code = await tryConnection();
+          if (code) {
+            document.getElementById('remote-sync-icon').style.color = 'green';
+            createSyncModal(`<h2>Sync code: ${code}</h2>`, code);
+          }
+        } else {
+          createSyncModal(`<h2>Must be online to use this feature.</h2>`, '');
```
bogas04: Do we even want to show sync button when user is offline?

tsingh777: That is difficult since we don't poll or get an event. if the user is online and then loses the connection, we don't have a way of getting that info to hide/show the button. I think it better to check on click.

www/js/menu.js
tsingh777: not sure if its needed. already one in `tryConnection` and isOnline should handle errors as well.

www/js/search.js
```diff
@@ -740,7 +753,8 @@ module.exports = {
     // load verses for bani based on baniID and the length that user has decided
     banidb.loadBani(BaniID, baniLengthCols[baniLength]).then(rowsDb => {
       // create a unique shabadID for whole bani, and append it with length
-      const shabadID = `${rowsDb[0].Token || rowsDb[0].Bani.Token}-${baniLength}`;
+      const shabadID = `${rowsDb[0].Token || rowsDb[0].Bani.Token}-${baniLength}-${rowsDb[0].ID ||
```
tsingh777: @inderpreetsingh please review if this is fine. needed to get the bani id on line selection.

www/js/search.js
```diff
@@ -740,7 +753,9 @@ module.exports = {
     // load verses for bani based on baniID and the length that user has decided
     banidb.loadBani(BaniID, baniLengthCols[baniLength]).then(rowsDb => {
       // create a unique shabadID for whole bani, and append it with length
-      const shabadID = `${rowsDb[0].Token || rowsDb[0].Bani.Token}-${baniLength}`;
+      const shabadID = `${rowsDb[0].Token || rowsDb[0].Bani.Token}-${baniLength}-${
```
tsingh777: @inderpreetsingh please review if this is fine. needed to get the bani id on line selection.

www/js/menu.js
```diff
@@ -301,6 +304,68 @@ const announcementSlideButton = h(
...
+  'li',
+  h(
+    'a.remote-sync-button',
+    {
+      onclick: async () => {
+        const onlineVal = await isOnline();
+        if (onlineVal) {
+          const code = await tryConnection();
+          if (code) {
+            document.getElementById('remote-sync-icon').style.color = 'green';
```
inderpreetsingh: `nitpicking` We don't use html green anywhere else, maybe use the green we use otherwise?

www/js/menu.js
```diff
@@ -301,6 +304,68 @@ const announcementSlideButton = h(
...
+    syncModal.addFooterBtn('Copy', 'tingle-btn tingle-btn--primary', () => {
+      copy(code);
+      syncModal.close();
+      syncModal.destroy();
+    });
+    syncModal.addFooterBtn('End Session', 'tingle-btn tingle-btn--primary', () => {
+      onEnd(code);
+      syncModal.close();
+      syncModal.destroy();
+      document.getElementById('remote-sync-icon').style.color = '#424242';
```
inderpreetsingh: I think it would be better if we add a class here, and then manipulate that through css. What do you think?

It just gets a bit hard managing styles when they are inside JS. Especially if I want to add more css rules here.

www/js/search.js
```diff
@@ -1014,11 +1028,21 @@ module.exports = {
...
-    /*
     if (window.socket !== undefined) {
-      window.socket.emit('data', { shabadid: ShabadID, highlight: LineID });
+      let shabadIdsplit = [ShabadID];
+      if (typeof ShabadID === 'string') {
+        shabadIdsplit = ShabadID.split('-');
+      }
+
+      window.socket.emit('data', {
+        type: shabadIdsplit.length > 1 ? 'bani' : 'shabad',
```
inderpreetsingh: I am little bit confused. How do we differentiate between bani and a ceremony?

www/js/search.js
inderpreetsingh: Looks good so far.

www/js/menu.js
tsingh777: this is temporary based on the mocks @saintsoldierx had. I can change it if you'd like

www/js/search.js
tsingh777: at the moment ceremony doesn't get the "-" with the bani length and id. not sure if we are going to be supporting ceremony either. might need to test that to see if it breaks anything.

PR #645, opened 2019-05-28
Windows 10 Shortcut Legend

AkalUstat:
I can explain....
So on saturday night I had to do shortcut legend twice because the first time it didn't work. 
The second time I was concerned with mac (trying to see if it worked) and it seemed that I forgot windows. 
Also I am adding this from a chromebook, at school, from github.com so I am unsure if this has any errors.

### DIFF
--- a/LEGEND.md
+++ b/LEGEND.md
@@ -14,17 +14,19 @@
 
 ### <a name="slides">Slide Shortcuts</a>
 
-#### Mac:
+#### MacOS:
 
 Cmd + e = empty slide
+Cmd + 1 = Waheguru Slide
 
 #### Windows:
 
 Ctrl + e = empty slide
+Ctrl + 1 = Waheguru Slide
 
 ### <a name="others">Others</a>
 
-#### Mac:
+#### MacOS:
 
 Cmd + q = Close app (shut down app)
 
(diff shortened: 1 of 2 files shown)


LEGEND.md
```diff
@@ -14,17 +14,19 @@
 
 ### <a name="slides">Slide Shortcuts</a>
 
-#### Mac:
+#### MacOS:
```
maneetpaul: macOS is the correct spelling/capitalization

LEGEND.md
```diff
@@ -14,17 +14,19 @@
...
 
 Cmd + e = empty slide
+Cmd + 1 = Waheguru Slide
```
maneetpaul: Each of these should be bullet points rather than being on a single line

LEGEND.md
```diff
@@ -14,17 +14,19 @@
...
 
 #### Windows:
 
 Ctrl + e = empty slide
+Ctrl + 1 = Waheguru Slide
 
 ### <a name="others">Others</a>
```
maneetpaul: I don't think we need to include a list of all the Windows/Mac shortcuts that are commonly used outside of STTM. We should keep this to just STTM keyboard shortcuts to prevent having a massive list.

maneetpaul:
Also @AkalUstat  Cmd/Ctrl + E is already there as a shortcut for displaying a blank slide.

PR #651, opened 2019-05-29
Update Vishraam Settings name

tarunsingh5:
updated the settings name for vishraams to say BaniDB living

### DIFF
--- a/www/js/settings.json
+++ b/www/js/settings.json
@@ -124,7 +124,7 @@
             "options": {
               "sttm2": "STTM 2",
               "igurbani": "iGurbani",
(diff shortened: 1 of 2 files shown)


www/js/settings.json
```diff
@@ -124,7 +124,7 @@
             "options": {
               "sttm2": "STTM 2",
               "igurbani": "iGurbani",
-              "sttm": "STTM Living"
+              "sttm": "BaniDB Living"
```
tsingh777: Needs to be updated to `HELP.md` as well

tarunsingh5: done

PR #662, opened 2019-06-08
Retains Shabad history

### LINKED ISSUE #329: History should retain Shabad status
All read lines should remain checked when clicking back to a Shabad and last selected line should come up. This is useful when switching back from a Parmaan.

Should be implemented along with Redux

inderpreetsingh:
https://youtu.be/T5ZvWAEwZS8

Fixes #329 

I know we wanted to do it via redux, so I am adding this PR here just as a quick alternative. Just in case we want to have a working version before we have redux refactor. So feel free to not merge this depending upon our priorities.

### DIFF
--- a/www/js/search.js
+++ b/www/js/search.js
@@ -22,6 +22,7 @@ const allowedKeys = [
   46, // Delete
 ];
 const sessionList = [];
+const sessionStatesList = {};
 const currentShabad = [];
 const kbPages = [];
 let currentMeta = {};
@@ -584,7 +585,7 @@ module.exports = {
       h(
         'a.panktee.current',
         {
-          onclick: ev => this.clickSession(ev, ShabadID, LineID),
+          onclick: ev => this.clickSession(ev, ShabadID),
         },
         Line.Gurmukhi,
       ),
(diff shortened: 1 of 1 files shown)


www/js/search.js
```diff
@@ -971,6 +988,8 @@ module.exports = {
       Array.from(lines).forEach(el => el.classList.remove('current'));
       // Add 'current' and 'seen-check' to selected Panktee
       $panktee.classList.add('current', 'seen_check');
+      shabadState.seenPanktees.add(LineID);
```
tsingh777: breaks ceremonies 
image

www/js/search.js
```diff
@@ -918,12 +928,17 @@ module.exports = {
...
+        delete sessionStatesList[shabadID];
+      });
     }
   },
 
-  clickSession(e, ShabadID, LineID) {
+  clickSession(e, ShabadID) {
     const $panktee = e.target;
-    this.loadShabad(ShabadID, LineID);
+    const [resumePanktee] = sessionStatesList[ShabadID];
```
tsingh777: not destructured correctly. should be `const { resumePanktee } = sessionStatesList[ShabadID];`

www/js/search.js
```diff
@@ -770,15 +772,22 @@ module.exports = {
       return false;
     }
 
+    let seenClasses = '';
+    const shabadState = sessionStatesList[line.ShabadID];
+
+    if (line.mainLine && !mainLineExists) {
+      seenClasses = '.current.main.seen_check';
+    } else if (shabadState.seenPanktees.has(line.ID)) {
```
tsingh777: breaks sundar gutka 
image

Issue #335, opened 2018-10-04
"Select Cast Device" Dialogue Box


maneetpaul:
When you press the cast button when in presenter view, the dialogue box shows up in the output panel really small instead of the entire window. 

![screen shot 2018-10-04 at 3 26 56 pm](https://user-images.githubusercontent.com/10658257/46497944-fcf4a900-c7e9-11e8-80fc-768874631883.png)

PR #665, opened 2019-06-12
Dhan Guru ______ shortcuts + simplified shortcuts

AkalUstat:
- [x] simplify existing shortcuts
- [x] clear search bar when using search bar shortcut
- [x] finish dhan guru shortcuts
- [x] update LEGEND.md with a table
- [x] create modal for slides
    - add dhan guru buttons (and give them functionality) 
    - add announcement tab
- [x]  comment code
https://gph.is/g/Zx1VWnq


inderpreetsingh:
<3

tsingh777:
> fix chromecast modal in viewer (#335)

do this as a separate pr

maneetpaul:
Any reason why we can't do CMD+Shift+1, 2, 3 etc, rather than Shift+G? Shift + G seems a little unnatural for a keyboard shortcut.

AkalUstat:
> Any reason why we can't do CMD+Shift+1, 2, 3 etc, rather than Shift+G? Shift + G seems a little unnatural for a keyboard shortcut.

Cmd + shift + 1 gets activiated by cmd +1 . Same with cmd + fn + 1

www/js/index.js
```diff
@@ -22,38 +22,30 @@ function escKey() {
...
-}
-function searchBarShortcut() {
-  search.$search.focus();
-}
+const slideShortcuts = {
+  waheguru: () => global.controller.sendText('vwihgurU', true),
+  empty: () => global.controller.sendText(' '),
+  moolMantra: () => {
+    global.controller.sendText(
+      '<> siq nwmu krqw purKu inrBau inrvYru Akwl mUriq AjUnI sYBM gur pRswid ]',
```
tsingh777: pulls these strings out in to a file just for constants and require them. They are probably used in multiple places so its better to keep it DRY(**D**ont**R**epeat**Y**ourself).

www/js/index.js
```diff
@@ -136,19 +128,50 @@ function findLine(e) {
     lineFound.click();
   }
 }
-
+const dhanguruStrings = [
+  'DMn gurU nwnk dyv jI',
```
tsingh777: same

AkalUstat:
> the keys for the dhan guru slides are a little awkward to press in my opinion. I could be wrong. @maneetpaul thoughts?

Yea i agree. However theyre the only combinations i could find that werent activated by mc + 1 ( the other shortcuts). Its a mousetrap problem

www/js/index.js
AkalUstat: Would u recommend a slideStrings.json or something because some of these are using in things like the toolbar

maneetpaul:
> > the keys for the dhan guru slides are a little awkward to press in my opinion. I could be wrong. @maneetpaul thoughts?
> 
> Yea i agree. However theyre the only combinations i could find that werent activated by mc + 1 ( the other shortcuts). Its a mousetrap problem

I rather we have it so CMD/CTRL + 1, 2, 3, 4 equate to the Dhan Guru slides (because those numbered keys make the most sense for all the Gurus) and we find different shortcuts for the Waheguru, Mool Mantar, Blank, etc. slides. Thoughts?

Maybe we can look into F1, F2, F3 etc for the Waheguru Slide, Mool Mantar, Blank, etc?

www/js/index.js
tsingh777: doesn't have to be a `.json` file.  `slideStrings.js` and in it do
```
export const slideString = {
 moolMantar: <> siq nwmu krqw purKu inrBau inrvYru Akwl mUriq AjUnI sYBM gur pRswid ]
 ...
}
```

and then use it in index.js like 
```
const slideString = require('./slideStrings');

...
global.controller.sendText(
      slideString.moolMantar,
      true,
    );
...
```

AkalUstat:
that should work

maneetpaul:
@AkalUstat - can we fix these shortctus in the next day or two so they can be ready for release in 5.2?

www/js/index.js
AkalUstat: This seems to break everything. It doesn't understand the export for some reason

AkalUstat: Im just gonna do a json for now

AkalUstat:
@maneetpaul can we do fn + # for dhan guru slides instead?

www/js/index.js
```diff
@@ -136,19 +127,33 @@ function findLine(e) {
...
+  Mousetrap.bindGlobal('fn+5', interfaceShortcuts.help);
+  Mousetrap.bindGlobal('fn+6', interfaceShortcuts.legend);
+  Mousetrap.bindGlobal('mod+/', interfaceShortcuts.searchBar);
+
+  for (let g = 0; g <= dhanGuruText.length - 1; g += 1) {
+    if (g <= 8) {
+      Mousetrap.bindGlobal([`mod+${g + 1}`], () =>
+        global.controller.sendText(dhanGuruText[g], true),
+      );
+      // eslint-disable-next-line eqeqeq
```
tsingh777: remove this and use `===`  See explanation: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness

www/js/index.js
```diff
@@ -136,19 +127,33 @@ function findLine(e) {
...
+    } else if (g == 9) {
+      Mousetrap.bindGlobal(['mod+0'], () => global.controller.sendText(dhanGuruText[g], true));
+      // eslint-disable-next-line eqeqeq
```
tsingh777: remove this and use `===`  See explanation: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness

www/js/slide-strings.js
```diff
@@ -0,0 +1,22 @@
+/* eslint-disable import/prefer-default-export */
```
tsingh777: remove this line

LEGEND.md
```diff
@@ -4,48 +4,88 @@
...
 
-- Cmd + q = Close app (shut down app)
-- Cmd w = Close app (shut down app)
-- Cmd + u = check for updates
-- Cmd + h = minimize window
-- Cmd + m = minimize window
+| Combination | Description       |
+| ----------- | ----------------- |
+| cmd + q     | close app         |
+| cmd + q     | close app         |
```
tsingh777: duplicate line?

LEGEND.md
```diff
@@ -4,48 +4,88 @@
...
+| Fn + 3     | Empty Slide                     |
+| Fn + 4     | Anand Sahib (Bhog)              |
+| Fn + 5     | Help Guide                      |
+| Fn + 6     | Shortcut Legend                 |
+| Cmd + /     | Jump to search bar              |
+| ----------- | -----------                     |
+| cmd + 1      | Dhan Guru Nanak Dev Ji          |
+| cmd + 2      | Dhan Guru Angad Dev Ji          |
+| cmd + 3      | Dhan Guru Amar Das Ji           |
+| cmd + 4      | Dhan Guru Raam Das Ji           |
```
tsingh777: Typically spelled as Guru Ram Das Ji with 1 `a` no?

LEGEND.md
```diff
@@ -4,48 +4,88 @@
...
-- Ctrl + 3 = Empty Slide
-- Ctrl + 4 = Anand Sahib (Bhog)
-- Ctrl + 5 = Help Guide
-- Ctrl + 6 = Shortcut Legend
-- Cmd + / = Jump to search
+#### <a name="macos-slides">macOS:</a>:
+
+| Combination | Description                     |
+| ----------- | ------------------------------- |
+| Fn + 1     | Waheguru Slide                  |
```
tsingh777: lets keep the casing for the `Fn` button consistent for mac and pc. probably want to make these `fn`

LEGEND.md
```diff
@@ -4,48 +4,88 @@
...
+| cmd + 8      | Dhan Guru Har Krishan Sahib Ji  |
+| cmd + 9      | Dhan Guru Teg Bahadur Sahib Ji  |
+| cmd + 0      | Dhan Guru Gobind Singh Sahib Ji |
+| cmd + -      | Dhan Guru Granth Sahib Ji       |
+
+#### <a name="windows-slides">Windows:</a>
+
+| Combination | Description                     |
+| ----------- | ------------------------------- |
+| fn + 1    | Waheguru Slide                  |
```
tsingh777: will this be an issue for windows keyboards that don't have a fn key?

AkalUstat: Probably

LEGEND.md
```diff
@@ -4,48 +4,67 @@
...
+| ----------- | ----------------------------------------- |
+| ctrl + 1    | Waheguru Slide                            |
+| ctrl + 2    | Mool Mantra Slide                         |
+| ctrl + 3    | Empty Slide                               |
+| ctrl + 4    | Anand Sahib (Bhog)                        |
+| ctrl + 5    | Help Guide                                |
+| ctrl + 6    | Shortcut Legend                           |
+| ctrl + /    | Jump to search bar                        |
+| ----------- | -----------                               |
+| ctrl + g    | Add Dhan Guru**\_** slide or announcement |
```
tsingh777: why the `**\_**` ?

www/js/index.js
```diff
@@ -16,44 +18,34 @@ const analytics = remote.getGlobal('analytics');
 /* const Settings = require('../../js/settings');
 const settings = new Settings(platform.store); */
 
+// const dhanGuruText = strings.slideStrings.dhanguruStrings;
```
tsingh777: remove

www/js/insert-slide.js
```diff
@@ -0,0 +1,188 @@
...
+    analytics.trackEvent('display', 'announcement-slide');
+    const isGurmukhi = document.querySelector('#modal-ann-lang').checked;
+    const announcementText = sanitizeHtml(document.querySelector('.modal-ann-box').innerHTML, {
+      allowedTags,
+    });
+    global.controller.sendText(announcementText, isGurmukhi);
+    document.querySelector('.modal-ann-box').innerHTML = '';
+    modal.close();
+  } else {
+    modal.close();
```
tsingh777: can move this after the `if` and just write it once for both cases and not have an `else` .
```
if (isAnnouncementTab) {
   ....
} 
modal.close();
```

www/js/insert-slide.js
```diff
@@ -0,0 +1,188 @@
...
+});
+
+// slide modal  body
+
+// title
+const slideHeader = '<h1 class = "modalTitle"> <center> Insert Slide </center></h1>';
+// section title
+const slideText = '<h2 class ="dhanguru">Dhan Guru:</h2>';
+// button group
+const buttons = '<center><div class="btn-group" id = "btn-group">'.concat(
```
tsingh777: center is an obsolete tag https://www.w3schools.com/tags/tag_center.asp

prefer if you accomplish this with css.

tsingh777: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/concat#Performance something to consider. performance should be negligible in this case but your call if you want to change it.

www/js/insert-slide.js
```diff
@@ -0,0 +1,188 @@
...
+ * sets the onclick val for the slider
+ * appends class gurmukhi to the text box (changes font family to gurbaniakhar in css file)
+ * changes placeholder text
+ */
+function setLangSliderVal() {
+  const slider = document.querySelector('#modal-ann-lang');
+  slider.onclick = () => {
+    // is the slider checkd?
+    const isGurmukhi = slider.checked;
+    const placeholderText = isGurmukhi ? 'GoSxw ie`Qy ilKo ...' : 'Add announcement text here ..';
```
tsingh777: pull from strings file

www/js/insert-slide.js
```diff
@@ -0,0 +1,188 @@
...
+}
+/**
+ * sets the onclick val for the slider
+ * appends class gurmukhi to the text box (changes font family to gurbaniakhar in css file)
+ * changes placeholder text
+ */
+function setLangSliderVal() {
+  const slider = document.querySelector('#modal-ann-lang');
+  slider.onclick = () => {
+    // is the slider checkd?
```
tsingh777: checked

www/js/index.js
```diff
@@ -136,19 +128,24 @@ function findLine(e) {
...
+  Mousetrap.bindGlobal(['mod+2'], slideShortcuts.moolMantra);
+  Mousetrap.bindGlobal(['mod+3'], slideShortcuts.empty);
 
-  Mousetrap.bindGlobal(['command+/', 'ctrl+/'], searchBarShortcut);
+  Mousetrap.bindGlobal(['mod+4'], ceremonyShortcuts.anandSahibBhog);
+
+  Mousetrap.bindGlobal('mod+5', interfaceShortcuts.help);
+  Mousetrap.bindGlobal('mod+6', interfaceShortcuts.legend);
+  Mousetrap.bindGlobal('mod+/', interfaceShortcuts.searchBar);
+  Mousetrap.bindGlobal('mod+g', () => {
```
tsingh777: @saintsoldierx @maneetpaul please review if this is what we are going for in terms of UX. 
image
image

www/js/index.js
```diff
@@ -16,44 +18,34 @@ const analytics = remote.getGlobal('analytics');
...
+    }),
+};
+const interfaceShortcuts = {
+  help: () => main.openSecondaryWindow('helpWindow'),
+  legend: () => main.openSecondaryWindow('shortcutLegend'),
+  searchBar: () => {
+    search.$search.focus();
+    search.$search.value = '';
+  },
+};
```
tsingh777: 👍

www/js/insert-slide.js
```diff
@@ -0,0 +1,188 @@
+const sanitizeHtml = require('sanitize-html');
+const electron = require('electron');
+
+const analytics = electron.remote.getGlobal('analytics');
+const tingle = require('./vendor/tingle');
+const strings = require('./strings');
+
+// allowed html tags inside announcement
+const allowedTags = [
```
tsingh777: lot of duplicate code in this file from menu. eventually would like that to be refactored to keep it [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)

AkalUstat: I can move this to the strings file maybe?

www/js/index.js
maneetpaul: This looks great. I would like it if we can use the arrow keys to navigate through everything so you don't need to use your mouse. Also, what shows up if you open "Slides" ?

maneetpaul: Also, is the shortcut to open this CMD + G?

AkalUstat: 1. not sure
2. yes

PR #666, opened 2019-06-13
History support 2

inderpreetsingh:
Retains shabad status (seen panktees, the main panktee etc) in shabad, bani and ceremonies.

Doing it separate from main history support PR so that if needed, this can be released separately. This is branched out from historySupport branch, so please merge it after `historySupport` is merged.

### DIFF
--- a/www/js/viewer.js
+++ b/www/js/viewer.js
@@ -87,11 +87,13 @@ const castShabadLine = lineID => {
     });
     castToReceiver();
 
-    const activeSlide = document.querySelector('.deck.active .slide.active').children;
-    Array.prototype.forEach.call(activeSlide, element => {
-      const icons = iconsetHtml(`icons-${element.classList[0]}`, element.innerHTML);
-      if (icons) document.querySelector('.viewer-controls').appendChild(icons);
-    });
+    const activeSlide = document.querySelector('.deck.active .slide.active');
+    if (activeSlide) {
+      Array.prototype.forEach.call(activeSlide.children, element => {
+        const icons = iconsetHtml(`icons-${element.classList[0]}`, element.innerHTML);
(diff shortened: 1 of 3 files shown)


tarunsingh5:
so excited by this

inderpreetsingh:
> if we add this, we should also have a way of clearing which lines have been seen, maybe when they search for it again?, because our keyboard shortcut for traversing with spacebar depends on the lines that have been seen.

Sure. @maneetpaul veerji how do we want to scope it?

maneetpaul:
Would we be able to ship this as is without breaking the spacebar shortcut? 

I agree with the logic that if the shabad is searched for / opened again (rather than being opened from the history panel) than it should have the checkmarks reset.

tsingh777:
> Would we be able to ship this as is without breaking the spacebar shortcut?

correct

www/js/viewer.js
```diff
@@ -87,11 +87,13 @@ const castShabadLine = lineID => {
...
-    const activeSlide = document.querySelector('.deck.active .slide.active').children;
-    Array.prototype.forEach.call(activeSlide, element => {
-      const icons = iconsetHtml(`icons-${element.classList[0]}`, element.innerHTML);
-      if (icons) document.querySelector('.viewer-controls').appendChild(icons);
-    });
+    const activeSlide = document.querySelector('.deck.active .slide.active');
+    if (activeSlide) {
+      Array.prototype.forEach.call(activeSlide.children, element => {
+        const icons = iconsetHtml(`icons-${element.classList[0]}`, element.innerHTML);
+        if (icons) document.querySelector('.viewer-controls').appendChild(icons);
```
tsingh777: add curly braces

PR #674, opened 2019-06-22
Viewer window edits

### LINKED ISSUE #671: Toggling Presenter View should reactivate viewer in second display
When you turn presenter view **off** - it should close the viewer that is typically being shown on the second display. When you turn it back **on**, it should throw the viewer back on the secondary display and make it the active window on that display. 

Currently, if something else becomes the active window on the secondary display, there is no way to get the viewer back up without closing and relaunching STTM

inderpreetsingh:
Fixes #671 

It hides the viewer window when presenter view is off, and brings it back on when the presenter view is switched back on.

### DIFF
--- a/app.js
+++ b/app.js
@@ -459,6 +459,18 @@ ipcMain.on('show-text', (event, arg) => {
   }
 });
 
+ipcMain.on('presenter-view', (event, arg) => {
+  if (viewerWindow) {
+    if (!arg) {
+      viewerWindow.setFullScreen(false);
+      viewerWindow.minimize();
+    } else {
+      viewerWindow.maximize();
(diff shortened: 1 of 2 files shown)


inderpreetsingh:
@maneetpaul veerji, I did not understand what "Some other window is active mean" but I reproduced it using the following steps. Please let me know, if I understood the bug correctly.

1) Open sttm with external display attached.
2) From external display minimize the viewer window
3) Open some other window, and move it to full screen
4) Now you cant bring back the viewer window automatically (can still bring it back in my case (macOS) from the minimized windows)

This PR fixes that, every time you switch presenter view it toggles (hides/shows) the viewer window too.

Issue #574, opened 2019-04-11
"Im feeling lucky" for desktop


tarunsingh5:
when you are searching and there is only one result pressing enter should load that shabad
<img width="780" alt="Screen Shot 2019-04-11 at 7 55 28 AM" src="https://user-images.githubusercontent.com/35271830/55967514-4c9bad80-5c2f-11e9-8e3c-d9dc248bda24.png">

tarunsingh5:
there's no "design" needed its just that when you press enter and there is only one search result it should load that shabad

PR #678, opened 2019-06-28
I'm Feeling Lucky Feature

AkalUstat:
~~If there is a singular result, clicking enter (while the search bar is still in focus) will open that shabad.~~
#574 
**Changes:** this will now open the first shabad in list, no matter how many results pop up as suggested by @maneetpaul.


www/js/index.js
```diff
@@ -136,6 +136,12 @@ function findLine(e) {
     lineFound.click();
   }
 }
+function feelingLucky() {
+  if (search.$results.childNodes.length === 1) {
```
tsingh777: i wouldn't call this a "feeling lucky". feeling lucky would be if `search.$results.childNodes.length` is `>=1`

www/js/index.js
```diff
@@ -136,6 +136,12 @@ function findLine(e) {
...
+    const singleResult = document.getElementsByClassName('panktee search-result');
```
tsingh777: I would do `document.querySelector('#results .search-result').click();`  incase we use that class somewhere else in the future, better to get the element with the idea and search within that.

Issue #675, opened 2019-06-24
Bani Overlay having issues with Ceremonies


maneetpaul:
Similar to #643 - when going through the Anand Karaj ceremony deck while using Bani Overlay, it will randomly show the Anand Karaj title slide in the overlay throughout the ceremony deck as you go through it. 

Check out the video below around 1:43:35:
https://youtu.be/6vesbIfMLXk?t=6215

PR #691, opened 2019-07-02
Fix issue with ceremony showing first line in overlay after paging

tsingh777:
should fix https://github.com/KhalisFoundation/sttm-desktop/issues/675

### DIFF
--- a/www/js/controller.js
+++ b/www/js/controller.js
@@ -554,7 +554,7 @@ module.exports = {
     if (document.body.classList.contains('livefeed')) {
       showLinePayload.live = true;
     }
-
+    // when paging, the first line gets loaded again. this makes sure obs shows the correct line.
     if (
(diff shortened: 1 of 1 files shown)


www/js/controller.js
```diff
@@ -554,7 +554,11 @@ module.exports = {
     if (document.body.classList.contains('livefeed')) {
       showLinePayload.live = true;
     }
-    if (start === 0 || start === undefined || mode === 'append') {
+
+    if (
+      (start === 0 || start === undefined || mode === 'append') &&
+      !(showLinePayload.Line.sessionKey.indexOf('ceremony') > -1 && mode === 'append' && start > 0)
```
inderpreetsingh: Maybe add a comment explaining what this does?

tsingh777: done.

PR #694, opened 2019-07-05
Increment version to 5.2.2

tsingh777:


maneetpaul:
We'll use 5.2.2 for now. But if #665 is completed this will change to 5.3

PR #707, opened 2019-07-12
Update Beta

maneetpaul:


maneetpaul:
Let's hold off on merging this until we fix #710

Issue #687, opened 2019-07-02
Anand Karaj and Anand Sahib don't appear in history on Win7


maneetpaul:
This only applies to the Anand Karaj ceremony and the Anand Sahib (Bhog) shortcut. They don't appear in the history panel correctly as other Bani's / Shabads do. Only verified this on Windows 7. All other Banis / Shabads work fine. 

<img width="695" alt="Screen Shot 2019-07-02 at 12 48 25 PM" src="https://user-images.githubusercontent.com/10658257/60531343-c68c8180-9cc8-11e9-8356-113a50523e6c.png">

PR #719, opened 2019-07-18
Adds name of ceremony to the ceremony object

inderpreetsingh:
Screenshot 2019-07-18 at 11 56 03 AM


maneetpaul:
I believe this is to fix #687

PR #723, opened 2019-07-23
Port overlay

inderpreetsingh:
https://youtu.be/LZeIU_xJrzo

Now the search for ports happen when you open window and click open the toggle.

### DIFF
--- a/app.js
+++ b/app.js
@@ -31,7 +31,7 @@ const store = new Store({
 
 const appVersion = app.getVersion();
 
-const { overlayCast } = store.getUserPref('app.overlay-cast');
+const overlayCast = store.getUserPref('app.overlay-cast');
 
 let mainWindow;
 let viewerWindow = false;
--- a/app.js
+++ b/app.js
@@ -24,33 +24,14 @@ expressApp.use(express.static(path.join(__dirname, 'www', 'obs')));
 
 const { app, BrowserWindow, dialog, ipcMain } = electron;
 
-op.find(
-  {
-    // Re: http://www.sikhiwiki.org/index.php/Gurgadi
-    ports: [1397, 1469, 1539, 1552, 1574, 1581, 1606, 1644, 1661, 1665, 1675, 1708],
-    count: 1,
-  },
-  (err, port) => {
-    if (err) {
-      dialog.showErrorBox(
-        'Overlay Error',
-        'No free ports available. Close other applications and Reboot the machine',
-      );
-      app.exit(-1);
-      return;
-    }
-    global.overlayPort = port;
-    // console.log(`Overlay Port No ${port}`);
-    http.listen(port);
-  },
-);
-
 const store = new Store({
   configName: 'user-preferences',
   defaults: defaultPrefs,
 });
 
 const appVersion = app.getVersion();
+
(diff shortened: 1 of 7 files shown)


www/js/overlay.js
```diff
@@ -13,8 +13,12 @@ const { store } = remote.require('./app');
 const analytics = remote.getGlobal('analytics');
 
 const { overlayVars } = store.get('obs').overlayPrefs;
+let overlayCast = store.getUserPref('app.overlay-cast');
 
-const url = `http://${host}:${overlayPort}/`;
+const controlPanel = document.querySelector('.control-panel');
+const webview = document.createElement('webview');
+
+let url = `http://${host}:${overlayPort}/`;
```
tsingh777: copy button's url doesn't get updated when initially off

app.js
```diff
@@ -301,6 +282,41 @@ const shouldQuit = app.makeSingleInstance(() => {
...
+      http.listen(port);
+    },
+  );
+};
+
+ipcMain.on('toggle-obs-cast', (event, arg) => {
+  if (arg) {
+    searchPorts();
+  } else {
+    http.close();
```
tsingh777: Doesn't look overlay actually turns off. updating lines still updates the webpage.

www/js/overlay.js
```diff
@@ -309,11 +339,12 @@ const themeSwatchFactory = themeOptions => {
 
 themeSelector.appendChild(h('div.theme-selector-header', 'Presets'));
 
+document.body.classList.toggle('overlay-off', !overlayCast);
```
tsingh777: the buttons still work when toggle is off

app.js
inderpreetsingh: Yes, I have shared about that in `sttm-desktop-dev`, please take a look. 🙏 That's the default behaviour of http.close(). It won't kill existing connections but prevent new connections.

www/js/overlay.js
inderpreetsingh: Yes, but it gets updated as soon as we click the on/off switch.

www/js/overlay.js
inderpreetsingh: Should be fixed now.

www/src/scss/_overlay.scss
```diff
@@ -157,6 +157,52 @@ input[type='color'] {
   width: 60px;
 }
 
+.overlay-off {
+  .cp-icon,
+  .setting-label {
+    color: #ddd;
+    cursor: not-allowed;
+    pointer-events: none;
```
tsingh777: minor thing: larivaar/padched button and copy url button still seem to work when disabled.

www/js/overlay.js
```diff
@@ -195,6 +201,31 @@ const toggleLarivaar = h(
   h('div.setting-label', `Use ${overlayVars.overlayLarivaar ? 'Padched' : 'Larivaar'}`),
 );
 
+const toggleCast = h(
+  'div.input-wrap.cast-toggle',
+  {
+    onclick: evt => {
```
tsingh777: perhaps we should also send empty slide if toggle is being turned off? doesn't have to be this PR.

PR #735, opened 2019-08-10
Increment to 5.3.3

maneetpaul:


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "productName": "SikhiToTheMax",
   "name": "sttm-desktop",
-  "version": "5.3.2",
+  "version": "6.0.0",
```
tsingh777: should we do 5.3.3 for that overlay bug?

maneetpaul: Agreed.

PR #752, opened 2019-08-29
adds setting to toggle announcement in bani overlay

inderpreetsingh:
Screenshot 2019-08-29 at 7 39 24 PM

### DIFF
--- a/app.js
+++ b/app.js
@@ -465,7 +465,11 @@ ipcMain.on('show-text', (event, arg) => {
       Transliteration: '',
     },
   };
-  showLine(textLine);
+
+  const announcementOverlay = store.getUserPref('app.announcement-overlay');
+  if (arg.isAnnouncement && announcementOverlay) {
+    showLine(textLine);
+  }
 
   if (viewerWindow) {
     viewerWindow.webContents.send('show-text', arg);
--- a/app.js
+++ b/app.js
@@ -466,8 +466,19 @@ ipcMain.on('show-text', (event, arg) => {
     },
   };
 
+  const emptyLine = {
+    Line: {
+      Gurmukhi: '',
+      English: '',
+      Punjabi: '',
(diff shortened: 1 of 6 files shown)


maneetpaul:
Sorry to nitpick - can we capitalize the word "Gurmukhi" that is below "Bani Overlay"

app.js
```diff
@@ -465,7 +465,11 @@ ipcMain.on('show-text', (event, arg) => {
       Transliteration: '',
     },
   };
-  showLine(textLine);
+
+  const announcementOverlay = store.getUserPref('app.announcement-overlay');
```
tsingh777: need an entry in `defaults.json` ?

inderpreetsingh: It's already there.

www/js/insert-slide.js
```diff
@@ -4,6 +4,11 @@ const electron = require('electron');
 const analytics = electron.remote.getGlobal('analytics');
 const tingle = require('./vendor/tingle');
 const strings = require('./strings');
+const settings = require('./settings');
+
+const { store } = electron.remote.require('./app');
+
+let announcementOverlay = store.getUserPref('app.announcement-overlay');
```
tsingh777: something buggy with this. It only seems to retrieve the defaults value. The checkbox is checked everytime the modal is opened regardless of the value in the setting. ideally they should be in sync.

www/js/insert-slide.js
```diff
@@ -52,6 +57,17 @@ const langSlider = '<div class="lang-switch">'.concat(
   '</div>',
   '</td>',
   '</tr>',
+  '<tr>',
+  '<td>',
+  '<span class="lang-text">Show Announcement in Overlay</span>',
```
tsingh777: the 2 switches could use some `padding-right` imo. 
image

PR #775, opened 2019-09-26
Update modal instructions text

### LINKED ISSUE #768: Sangat Sync: Update modal instructions text
Let’s slightly change the modal text: 

Enter this code on sttm.co/sync to follow along SikhiToTheMax on a mobile device (internet required). Press the _Present_ button to display the pairing code for the sangat.

maneetpaul:
Fixes #768

### DIFF
--- a/www/js/toolbar.js
+++ b/www/js/toolbar.js
@@ -88,7 +88,7 @@ const syncContent = h('div.sync-content', [
     h('div.sync-code-num', code),
     h(
       'div.sync-code-desc',
-      'Share this code with anyone using a mobile device* and they can open their browser, go to sttm.co/sync and enter the code above to follow along with the desktop app on their device',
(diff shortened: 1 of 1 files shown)


www/js/toolbar.js
```diff
@@ -88,7 +88,7 @@ const syncContent = h('div.sync-content', [
     h('div.sync-code-num', code),
     h(
       'div.sync-code-desc',
-      'Share this code with anyone using a mobile device* and they can open their browser, go to sttm.co/sync and enter the code above to follow along with the desktop app on their device',
+      'Enter this code on sttm.co/sync to follow along SikhiToTheMax on a mobile device (internet required). Press the "Present" button to display the pairing code for the sangat.',
```
tsingh777: Would it better to specify "any web browser (including mobile)" in stead of "mobile device" ? not sure if it would make it seem like this feature is only for mobile.

maneetpaul: Fair point!

PR #789, opened 2019-11-07
Use black Sync icon on light themes

maneetpaul:
Feel free to close this and replace it with another PR if this is not the correct way to fix this issue.


inderpreetsingh:
It's perfect, I would just remove the css rule about icon background from the styles.css so that it's in either in light base or in dark base and nowhere else.

PR #826, opened 2019-11-14
adds setting to change announcement text size

inderpreetsingh:
Announcement text has a font size now, that's different from gurbani and all.

You can change that setting in settings panel.

### DIFF
--- a/www/js/viewer.js
+++ b/www/js/viewer.js
@@ -104,6 +104,14 @@ const castText = (text, isGurmukhi) => {
   castCur.larivaar = text;
   castCur.isText = true;
   castToReceiver();
+
+  const activeSlide = document.querySelector('.deck.active .slide.active');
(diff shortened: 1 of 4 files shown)


www/js/viewer.js
```diff
@@ -428,12 +436,16 @@ const showText = (text, isGurmukhi = false) => {
...
   const $textIs = document.createElement('div');
-  $textIs.classList.add('gurbani');
+  $textIs.classList.add('announcements');
   if (isGurmukhi) {
     $textIs.classList.add('gurmukhi');
   }
   $textIs.innerHTML = text;
 
+  /* If slide is not empty, show quick tools */
+  const $vcToggleIcon = document.querySelector('.vc-toggle-icon');
```
tsingh777: what is vc?

Issue #829, opened 2019-11-14
Change default Announcement font size to 9 


maneetpaul:
Same default as Bani

Issue #830, opened 2019-11-14
Verify padding between lines in Mool Mantra 


maneetpaul:
<img width="778" alt="Screen Shot 2019-11-14 at 11 49 47 AM" src="https://user-images.githubusercontent.com/10658257/68878344-8c4eda00-06d5-11ea-9ae0-4c93a27d884b.png">

PR #831, opened 2019-11-14
Announcement text changes

inderpreetsingh:
Adds line height to announcement text

Changes default setting for announcement


maneetpaul:
Fixes #829 and #830

PR #858, opened 2019-11-26
removes the complex paging code

### LINKED ISSUE #835: Anand Karaj Ceremony stops displaying to secondary display after the 2nd laav
Similar to the issue we had with the last few lines of Anand Sahib not showing up on the second display. 

It shows in the preview viewer fine.

inderpreetsingh:
So, this was a bit weirder than I thought. Fixes #835 

I tried to fix it in bunch of ways, which ended up feeling like hacks. Eventually, I was able to find the root cause by going through our history.  We introduced this bug when we were trying to fix the paging problem in the overlay  (#740 #691 ) Later in history, we changed the way `sessions` works and also changed the id for the ceremony (not sure), I am suspecting that caused this bug to creep up.

I removed those complex if conditions and It did the trick. I thoroughly tested to make sure that the original bugs we fixed did not resurface. As far as my testing goes, it looks good. I have created a video of going through the ceremony (anand karaj) and going through a long bani (sukhmani sahib) on overlay and second display; It works fine for me. 🤞 

 https://youtu.be/EF_MLHkf7p4

@tsingh777 as you wrote that code, you might have more insights. Can you help me test this PR in the context of those bugs as well? May be give some test recipes? There's good chance that l might be missing something in understanding those bugs. 🙏

### DIFF
--- a/www/js/controller.js
+++ b/www/js/controller.js
@@ -555,18 +555,7 @@ module.exports = {
     if (document.body.classList.contains('livefeed')) {
       showLinePayload.live = true;
     }
-    // when paging, the first line gets loaded again. this makes sure obs shows the correct line.
-    if (
-      (start === 0 || start === undefined || mode === 'append') &&
-      !(
-        showLinePayload.Line.sessionKey &&
-        showLinePayload.Line.sessionKey.indexOf('ceremony') > -1 &&
-        mode === 'append' &&
-        start > 0
-      )
-    ) {
-      global.platform.ipc.send('show-line', showLinePayload);
-    }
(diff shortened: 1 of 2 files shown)


tsingh777:
Happy paths look okay. One error I'm seeing 

`/sttm-desktop/www/js/search.js:1085 Uncaught TypeError: Cannot read property 'seenPanktees' of undefined
    at Object.clickShabad (/Users/tsingh7/development/sttm-desktop/www/js/search.js:1085)`

this happens after a second screen is added. maybe we can add a null check but something seems off with that. Not able to reproduce this error in `dev`.

PR #859, opened 2019-11-26
adds default theme fallback in case theme does not exist

### LINKED ISSUE #851: Reset theme if Guru Nanak theme was last used theme
If you are on the Guru Nanak theme and update, it causes the UI to get all weird. Need to select a theme to fix it. 

If Guru Nanak theme was the last selected theme, it should change it to the light theme when updating.

inderpreetsingh:
Fixes #851

### DIFF
--- a/www/js/theme_editor.js
+++ b/www/js/theme_editor.js
@@ -8,7 +8,6 @@ const { remote } = require('electron');
 const readChunk = require('read-chunk');
 const imageType = require('image-type');
 
-const themes = require('./themes.json');
 const slash = require('./slash');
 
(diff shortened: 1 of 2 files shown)


www/js/theme_editor.js
```diff
@@ -298,6 +298,11 @@ module.exports = {
   init() {
     const themeOptions = document.querySelector('#custom-theme-options');
 
+    const currentTheme = themes.find(theme => theme.key === store.getUserPref('app.theme'));
```
tsingh777: Does this work? I think the store value needs to set before the render process is started (in `app.js`)

PR #898, opened 2019-12-19
Fixes devanagari typo and changes position of language settings

inderpreetsingh:
Fixes Devanagari typo and changes position of language settings. Did them together because doing them separately would have resulted in merge conflicts.


inderpreetsingh:
@maneetpaul it should be capital D - that commit came like a minute later. 🙏

PR #914, opened 2019-12-28
Add default settings to pass to chromecast and increment version

tsingh777:
https://github.com/KhalisFoundation/sttm-chromecast/pull/25 will catch issues in current prod. Need to confirm with @inderpreetsingh the source of truth for the language settings.

### DIFF
--- a/www/js/defaults.json
+++ b/www/js/defaults.json
@@ -33,6 +33,10 @@
       "vishraam": {
         "vishraam-options": "colored-words",
         "vishraam-source": "sttm"
+      },
+      "language-settings": {
+        "translation-language": "English",
+        "transliteration-language": "English"
       }
     },
     "slide-layout": {
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "productName": "SikhiToTheMax",
   "name": "sttm-desktop",
-  "version": "6.3.1",
+  "version": "6.3.2",
   "description": "The SikhiToTheMax desktop app",
   "main": "app.js",
   "scripts": {


inderpreetsingh:
looks good to me, will test it once on chromecast and approve. 🙏

www/js/defaults.json
```diff
@@ -33,6 +33,10 @@
       "vishraam": {
         "vishraam-options": "colored-words",
         "vishraam-source": "sttm"
+      },
+      "language-settings": {
```
tsingh777: @inderpreetsingh this is duplicated from the `slide-layout` preference. which is needed?

inderpreetsingh: Yes its supposed to only be in slide-layout and that is needed. We access the settings in codebase as `store.getUserPref('slide-layout.language-settings.translation-language');`

inderpreetsingh: I am sorry if I am missing something obvious but I am unable to understand why this change was made. Can you please write a line or two explaining the issue?

tsingh777: Chromecast was looking at the `pref.toolbar` language setting. https://github.com/KhalisFoundation/sttm-chromecast/blob/master/src/js/index.js#L62 

I updated to chromecast to look for `prefs['slide-layout']` if `pref.toolbar` isn't there. I don't know if there was a version that set `pref.toolbar` for language settings. If there isn't one then we can clean up chromecast to not look for `pref.toolbar` at all and decline this PR.

inderpreetsingh: I will look into that. Thank you for addressing the chromecast fix 🙏

Issue #915, opened 2020-01-01
Bani Overlay English translation broken


saintsoldierx:
When using Bani Overlay, it does not display the translations using any layout.

PR #918, opened 2020-01-03
Pull languages from correct user settings for overlay

tsingh777:
Should fix https://github.com/KhalisFoundation/sttm-desktop/issues/915

### DIFF
--- a/app.js
+++ b/app.js
@@ -262,9 +262,10 @@ const showLine = (line, socket = io) => {
   const overlayPrefs = store.get('obs');
   const lineWithSettings = line;
   lineWithSettings.languageSettings = {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "productName": "SikhiToTheMax",
   "name": "sttm-desktop",
-  "version": "6.3.1",
+  "version": "6.3.2",
```
maneetpaul: We incremented with the chromecast PR so this isn’t needed

inderpreetsingh:
Other than what @maneetpaul singh veerji said, rest is fine. :)

PR #922, opened 2020-01-05
Remove toolbar language duplication in default settings json

tsingh777:


AkalUstat:
Will break copy again

tsingh777:
@AkalUstat This won't. This was a default setting that wasn't needed.

PR #928, opened 2020-01-09
Adding admin code functionality to sync - One socket

inderpreetsingh:
https://www.youtube.com/watch?v=BSFt5ITQtac&feature=youtu.be

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "productName": "SikhiToTheMax",
   "name": "sttm-desktop",
-  "version": "7.0.0-inder",
+  "version": "7.0.0",
   "description": "The SikhiToTheMax desktop app",
   "main": "app.js",
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "productName": "SikhiToTheMax",
   "name": "sttm-desktop",
-  "version": "6.3.2",
+  "version": "7.0.0-inder",
   "description": "The SikhiToTheMax desktop app",
   "main": "app.js",
   "scripts": {
--- a/www/js/search.js
+++ b/www/js/search.js
@@ -316,7 +316,7 @@ function akhandPaatt() {
   // global.controller.clearAPV();
 }
 
-const currentShabadID = () => currentShabadID;
+const getCurrentShabadID = () => currentShabadID;
 
 module.exports = {
   currentShabad,
--- a/www/js/search.js
+++ b/www/js/search.js
@@ -320,7 +320,7 @@ const getCurrentShabadID = () => currentShabadID;
 
 module.exports = {
   currentShabad,
-  currentShabadID,
+  getCurrentShabadID,
   currentMeta,
   baniLengthCols,
 
--- a/www/js/search.js
+++ b/www/js/search.js
@@ -316,15 +316,11 @@ function akhandPaatt() {
   // global.controller.clearAPV();
 }
 
-const currentShabadState = () => ({
-  currentShabad,
-  currentShabadID,
-  sessionStatesList,
-});
+const currentShabadID = () => currentShabadID;
 
 module.exports = {
   currentShabad,
-  currentShabadState,
+  currentShabadID,
   currentMeta,
   baniLengthCols,
 
--- a/www/js/search.js
+++ b/www/js/search.js
@@ -22,6 +22,7 @@ const allowedKeys = [
 const sessionList = [];
 const sessionStatesList = {};
 const currentShabad = [];
+let currentShabadID = 0;
 const kbPages = [];
 let currentMeta = {};
 let newSearchTimeout;
@@ -315,8 +316,15 @@ function akhandPaatt() {
   // global.controller.clearAPV();
 }
 
+const currentShabadState = () => ({
+  currentShabad,
+  currentShabadID,
+  sessionStatesList,
+});
+
 module.exports = {
   currentShabad,
+  currentShabadState,
   currentMeta,
   baniLengthCols,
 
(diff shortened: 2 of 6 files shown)


maneetpaul:
The two notes we discussed on the call: 

- Reverse the way the toggle works currently. Code should be shown when the icon is clicked. The toggle should disable the code, and turning the toggle off should re-enable it
- Let's make both URLs **bold**

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "productName": "SikhiToTheMax",
   "name": "sttm-desktop",
-  "version": "6.3.2",
+  "version": "7.0.0-inder",
```
maneetpaul: Please remove the -inder

www/js/toolbar.js
```diff
@@ -82,58 +131,89 @@ const switchFactory = (id, label, inputId, clickEvent, defaultValue = true) =>
...
         },
-        'Present',
-      ),
-    ]),
+      },
+      'Present Code to Sangat',
+    ),
+  ),
+  syncItemFactory(
+    'Remote Control',
```
maneetpaul: Let's rename this to "Bani Controller" to stay in line with marketing

www/js/search.js
```diff
@@ -315,8 +316,15 @@ function akhandPaatt() {
   // global.controller.clearAPV();
 }
 
+const currentShabadState = () => ({
```
inderpreetsingh: Current shabad state can be reduced to just shabadID.

www/js/toolbar.js
```diff
@@ -45,18 +48,64 @@ const toggleOverlayUI = (toolbarItem, show) => {
   });
 };
 
+const setListeners = () => {
```
inderpreetsingh: May be: Should move setListeners to a separate file or in share-sync file.

www/js/search.js
```diff
@@ -315,8 +316,15 @@ function akhandPaatt() {
...
+  currentShabad,
+  currentShabadID,
+  sessionStatesList,
+});
+
 module.exports = {
   currentShabad,
+  currentShabadState,
```
inderpreetsingh: History functionality still needs to be tested properly.

www/js/toolbar.js
```diff
@@ -82,58 +131,89 @@ const switchFactory = (id, label, inputId, clickEvent, defaultValue = true) =>
...
+  ),
+]);
+
 const syncContent = h('div.sync-content', [
-  h('div.left-sync-content', [
+  h('div.sync-code-label', 'Your unique sync code is:'),
+  h('div.sync-code-num', code),
+  syncItemFactory(
+    'Sangat Sync',
+    'Allow the Sangat to visit <strong> sttm.co/sync </strong> to enter a code and view the Shabad being presented in real-time.',
```
inderpreetsingh: HTML doesn't work here. Need to fix.

www/src/scss/styles.scss
```diff
@@ -361,43 +361,66 @@ a {
...
     max-width: 600px;
   }
 
   .sync-content {
     align-items: center;
     border-radius: 5px;
     display: flex;
-    height: 160px;
-    justify-content: space-around;
+    flex-direction: column;
```
inderpreetsingh: may be move to separate sync file with variables for sync.

www/js/toolbar.js
```diff
@@ -45,18 +48,64 @@ const toggleOverlayUI = (toolbarItem, show) => {
   });
 };
 
+const setListeners = () => {
+  if (window.socket !== undefined) {
+    window.socket.on('data', data => {
+      const { search } = global.core.search;
+      const isPinCorrect = parseInt(data.pin, 10) === adminPin;
+
+      const loadShabad = (shabadId, verseId, gurmukhi) => {
```
inderpreetsingh: add comment here for why we need gurmukhi.

// We need gurmukhi here to add for history support. Yes, we can get the gurmukhi from desktop side as well, but I didn't want to mess with current history code. As that could lead to a lot of bugs.. which is already a minefield because of documented state management issues with current code.

www/js/toolbar.js
```diff
@@ -45,18 +48,64 @@ const toggleOverlayUI = (toolbarItem, show) => {
...
+        },
+        /* Coming soon
+        'bani' : global.core.search.loadBani(data.baniId, data.verseId); 
+        'ceremony' : global.core.search.loadCeremony(data.ceremonyId, data.verseId); 
+        */
+      };
+
+      // if its an event from web and not from desktop itself
+      if (data.host === 'sttm-web') {
+        listenerActions[isPinCorrect ? data.type : 'request-control'](data);
```
inderpreetsingh: Need to make  sure whether we need to test the pin on each request or not, may be once in starting is fine.

www/js/toolbar.js
tsingh777: Please do keep files small.

tsingh777: There's also an `addListener` function in share-sync that handlest the socket check.

www/js/toolbar.js
```diff
@@ -45,18 +48,64 @@ const toggleOverlayUI = (toolbarItem, show) => {
...
+          search.addToHistory(shabadId, verseId, gurmukhi);
+        }
+      };
+
+      const listenerActions = {
+        shabad: payload => loadShabad(payload.shabadId, payload.verseId, payload.gurmukhi),
+        text: payload =>
+          global.controller.sendText(payload.text, payload.isGurmukhi, payload.isAnnouncement),
+        'request-control': () => {
+          window.socket.emit('data', {
```
tsingh777: is this to tell web if the pin correct?

inderpreetsingh: @tsingh777 yes!

inderpreetsingh:
Screenshot 2020-01-15 at 6 20 08 PM

*Offline Support*

inderpreetsingh:
Screenshot 2020-01-15 at 6 30 56 PM

Issue #934, opened 2020-01-12
Live feed text files not working


saintsoldierx:
Clicking toggle button in settings for live feed text files is not working (Mac). I click the toggle button, popup comes up to select file save location, but toggle stays deactivated and the files are not generated.

PR #942, opened 2020-01-16
Fix overlay files not being written

tsingh777:
Fixes https://github.com/KhalisFoundation/sttm-desktop/issues/934

After node upgrade, callback function was required. https://nodejs.org/docs/latest-v10.x/api/fs.html#fs_fs_writefile_file_data_options_callback


inderpreetsingh:
May be we can add analytics in that callback as well?

Issue #916, opened 2020-01-01
Make Bani Overlay URL more visible


saintsoldierx:
1. Need a reset button to re-load the default settings.
2. Need a show URL button as well as copy. Sometimes user needs to show URL on screen to enter on another laptop for Bani Overlay.

maneetpaul:
@saintsoldierx if you hover over the copy icon it will display the URL. Is there more we should do there?

saintsoldierx:
I would include URL that is always displayed if possible? I keep forgetting about the hover and im sure others may do too. I would show it at the bottom, below the preview screen on Bani Overlay, small not large font.

saintsoldierx:
^ and put a copy button next to it or something, remove from the left side bar.

maneetpaul:
Rough mockup

![OverlayURL](https://user-images.githubusercontent.com/10658257/72365959-4c9d7200-36c7-11ea-858e-23e23615cc55.jpg)

PR #962, opened 2020-01-24
Move overlay url

AkalUstat:
Screen Shot 2020-01-23 at 6 16 10 PM

### DIFF
--- a/www/js/overlay.js
+++ b/www/js/overlay.js
@@ -14,6 +14,7 @@ let announcementOverlay = store.getUserPref('app.announcement-overlay');
 
 const controlPanel = document.querySelector('.control-panel');
 const textControls = document.querySelector('.text-controls');
+const urlContainer = document.querySelector('.url-container');
 const webview = document.createElement('webview');
 
 const getUrl = () => {
@@ -167,19 +168,6 @@ const resizeButtonFactory = (increaseFunc, decreaseFunc) =>
     ),
   );
 
-const copyURLButton = h(
-  'div.input-wrap',
-  {
-    title: `${url}`,
-    onclick: () => {
-      copy(url);
-      analytics.trackEvent('overlay', 'urlCopied', url);
-    },
-  },
-  h('div.export-btn', h('i.fa.fa-files-o.cp-icon')),
-  h('div.setting-label', 'Copy URL'),
-);
-
 const toggleLarivaar = h(
   'div.input-wrap',
   {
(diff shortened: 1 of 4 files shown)


AkalUstat:
fixes #916

maneetpaul:
@AkalUstat Can we add a little padding between the box and the URL so they arent so close together? 
Great job!

AkalUstat:
@maneetpaul bhai sahib ji i made some styling changes directly from gh web client so im not sure if  the css worked. If you could check that would be nice 🙏

AkalUstat:
Screen Shot 2020-01-24 at 4 03 04 PM
@saintsoldierx @maneetpaul

AkalUstat:
@maneetpaul @saintsoldierx 
Screen Shot 2020-01-25 at 6 37 45 PM

maneetpaul:
LGTM. A little padding on top and bottom as well would be nice but this works. 

@tsingh777 for code review

www/src/scss/_overlay.scss
```diff
@@ -61,6 +61,30 @@
     width: auto;
   }
 }
+.url-container {
+  display: inline-block;
```
tsingh777: not needed.

www/src/scss/_overlay.scss
```diff
@@ -61,6 +61,30 @@
...
+  margin-top: 1vh;
+  padding: 8px;
+
+  & > div {
+    display: inline-block;
+  }
+  .input-wrap {
+    align-items: center;
```
tsingh777: This is for `display: flex` you don't need it if your using inline-block everywhere. please remove.

www/src/scss/_overlay.scss
```diff
@@ -61,6 +61,30 @@
     width: auto;
   }
 }
+.url-container {
+  display: inline-block;
+  margin-top: 1vh;
+  padding: 8px;
```
tsingh777: this element should have either padding or margin (preferably margin). Let stick with one, use `margin: 9px` maybe?

www/js/overlay.js
```diff
@@ -278,12 +266,28 @@ const toggleAnnouncements = h(
   h('div.setting-label', 'Announcement'),
 );
 
+const urlText = h(
+  'div.copy-text',
+  h('text.url', {
+    innerText: `${url}`,
```
tsingh777: when overlay is initially off we get this:
image
When it gets turned on the URL doesn't update and it still has `undefined` as the port. 

Can you add an update to the url when overlay is turned on. Also when it is off, maybe hide it?

PR #963, opened 2020-01-29
adds larivaar assist functionality

inderpreetsingh:
Screenshot 2020-01-29 at 4 08 48 PM
Added larivaar assist functionality that overtakes vishraam styles when its on.

### DIFF
--- a/www/src/scss/themes/helpers/_theme-generator.scss
+++ b/www/src/scss/themes/helpers/_theme-generator.scss
@@ -45,6 +45,12 @@
         );
       }
     }
+
+    &.larivaar.larivaar-assist {
+      .gurbani .larivaar span:nth-of-type(even) {
+        color: $visraam-yamki;
+      }
+    }
   }
 }
 
--- a/www/src/scss/themes/helpers/_theme-generator.scss
+++ b/www/src/scss/themes/helpers/_theme-generator.scss
@@ -7,42 +7,47 @@
   $visraam-sources: ('sttm', 'sttm2', 'igurbani');
 
   @each $source in $visraam-sources {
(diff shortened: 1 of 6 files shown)


maneetpaul:
Will this work on Chromecast and is it compatible across dark/light themes?

inderpreetsingh:
@maneetpaul I have added theme support by using the theme respective yamki color.

inderpreetsingh:
I will send chromecast PR on chromecast repo.

www/src/scss/themes/helpers/_theme-generator.scss
```diff
@@ -45,6 +45,12 @@
         );
       }
     }
+
+    &.larivaar.larivaar-assist {
```
tsingh777: so larivaar assist will have the highest priority? would some messaging help so users don't get confused?

maneetpaul:
This should work similar to our vishraam options. Where Larivaar assist is only an available option if Larivaar is first turned on. If Larivaar is turned off, then assist is also automatically turned off and disabled. Once assist is turned off, vishraams come back as their previous state. 

That should eliminate any confusion.

ManjotS:
Is it possible to have vishraam supercede assist? So we can use both?

maneetpaul:
@ManjotS, for now, we are going to have it be one or another. We'll enhance this down the road to allow for both. I also want to see what usage is like for larivaar assist in our analytics.

ManjotS:
It will be low because:

- very few people know or care what larivaar is (as percentage of panth)
- Of those that know, a smaller portion care to use it
- Of those that know, some are against using it

inderpreetsingh:
https://youtu.be/75D_jcbLp8s - Larivaar options :)

PR #965, opened 2020-01-29
adds new teeka source

inderpreetsingh:
Adds 2  teeka sources in settings - Faridkot teeka & Sahib Singh teeka. 

I will send future PRs to get it supported on
- Chromecast view
- Bani overlay
- While Copying


maneetpaul:
As per Slack conversation with @ManjotS - we just need to replace the current source for now. Not add an additional option.

PR #1026, opened 2020-03-17
Reset overlay settings

inderpreetsingh:
Reset overlay settings button - that brings all the settings to default.


www/js/overlay.js
```diff
@@ -1,7 +1,9 @@
+/* eslint-disable import/no-unresolved */
 const electron = require('electron');
 const h = require('hyperscript');
 const ip = require('ip');
 const copy = require('copy-to-clipboard');
+const { obs: defaultPrefs } = require('./js/defaults.json');
```
tsingh777: is this path correct? this file is in the same directory as `defaults.json`

inderpreetsingh: Yes, the paths are referenced from `www` folder. So `./defaults.json` would give an error.

tsingh777:
this will probably have conflicts with https://github.com/KhalisFoundation/sttm-desktop/pull/1016 so lets get that in first?

www/js/overlay.js
```diff
@@ -1,7 +1,9 @@
+/* eslint-disable import/no-unresolved */
```
tsingh777: prefer if we do `// eslint-disable-next-line import/no-unresolved` before the line

PR #1035, opened 2020-03-24
adds larivaar assist support to overlay

inderpreetsingh:
https://youtu.be/3u59qYd_MR8

### DIFF
--- a/www/src/scss/obs/styles.scss
+++ b/www/src/scss/obs/styles.scss
@@ -51,6 +51,15 @@ body {
   position: relative;
   text-align: center;
   width: 100%;
+
+  span.normal-word,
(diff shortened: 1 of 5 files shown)


www/src/scss/obs/styles.scss
```diff
@@ -51,6 +51,15 @@ body {
...
   text-align: center;
   width: 100%;
+
+  span.normal-word,
+  span.light-word {
+    word-break: break-word;
+  }
+
+  span.light-word {
+    opacity: 0.5;
```
tsingh777: this seems low. for some color choice you can't see it. not sure if its worth optimizing though.

PR #1036, opened 2020-03-24
Adds support for fonts in overlay

inderpreetsingh:
Adds multiple font options in bani overlay window.
Screenshot 2020-03-25 at 1 06 44 AM


maneetpaul:
@inderpreetsingh is the window size reactive? Meaning, if the window is big enough, will the layout options appear on the same row as the other options?

www/js/overlay.js
```diff
@@ -176,6 +176,46 @@ const resizeButtonFactory = (increaseFunc, decreaseFunc) =>
...
+window.addEventListener('click', e => {
+  if (!e.target.classList.contains('select-value')) {
+    closeAllDropDowns();
+  } else {
+    toggleDropDown(e.target.parentElement);
+  }
+});
+
+const fontListFactory = (list, propName) => {
+  const lowerize = str => str.replace(/ /g, '').toLowerCase();
```
tsingh777: should probably define this outside of factory function.

www/src/scss/obs/styles.scss
```diff
@@ -1,3 +1,45 @@
...
+}
+
+@font-face {
+  font-family: 'caveat';
+  src: url('english/Caveat.ttf') format('truetype');
+}
+
+@font-face {
+  font-family: 'ptserif';
+  src: url('english/PT-Serif.ttf') format('truetype');
```
tsingh777: think this english folder is missing

www/js/overlay.js
```diff
@@ -176,6 +176,46 @@ const resizeButtonFactory = (increaseFunc, decreaseFunc) =>
...
+  const onClickOption = (e, font) => {
+    e.target.parentElement.parentElement.querySelector('.select-value').innerHTML = font;
+    overlayVars[propName] = lowerize(font);
+    savePrefs();
+  };
+  const options = list.map(font => h(`div.option`, { onclick: e => onClickOption(e, font) }, font));
+  return h(
+    'div.custom-select',
```
tsingh777: why didn't we just use a select element like we have in other parts of the app?

www/js/defaults.json
```diff
@@ -91,6 +104,8 @@
         "gurbaniTextColor": "#EAC445",
         "overlayLogo": true,
         "overlayLarivaar": false,
+        "translationFont": "muli",
+        "gurbaniFont": "gurbaniakhar",
```
tsingh777: why not just keep it the same as font name? that would remove need to transform the value when reading/writing.

www/js/overlay.js
```diff
@@ -176,6 +176,46 @@ const resizeButtonFactory = (increaseFunc, decreaseFunc) =>
...
+  if (!e.target.classList.contains('select-value')) {
+    closeAllDropDowns();
+  } else {
+    toggleDropDown(e.target.parentElement);
+  }
+});
+
+const fontListFactory = (list, propName) => {
+  const lowerize = str => str.replace(/ /g, '').toLowerCase();
+  const onClickOption = (e, font) => {
```
tsingh777: define outside of factory function.

www/js/defaults.json
tsingh777: Just wondering but can the bold ones be their own font family? The key is arbitrary so would it makes it easier for each option to correspond to its own font-face?

www/obs/index.html
```diff
@@ -89,6 +89,23 @@
         document.querySelector('.o-gurbani').style.color =
           overlayPrefs.overlayVars.gurbaniTextColor;
 
+        document.body.classList.remove(
```
tsingh777: is this duplicate? the same is done on line 109

www/js/overlay.js
inderpreetsingh: It wasn't looking as good here - and also needed some functionality which allows me to open the dropdown as drop-up depending on window size.

inderpreetsingh:
@maneetpaul yes it is now. :)

www/js/overlay.js
```diff
@@ -176,6 +176,48 @@ const resizeButtonFactory = (increaseFunc, decreaseFunc) =>
...
+  overlayVars[propName] = font;
+  savePrefs();
+};
+
+const fontListFactory = (list, propName) => {
+  const options = list.map(font =>
+    h(`div.option`, { onclick: e => fontSwitch(e, font, propName) }, font),
+  );
+  return h(
+    'div.custom-select',
```
tsingh777: could have probably extracted this component to separate file. can follow up on this later.

PR #1042, opened 2020-04-02
I18n. - Separate Strings 

### LINKED ISSUE #565: Refactor: Move all strings to a separate place (may be use i18n)
Our list of user-facing messages is constantly increasing. We should maintain one tone, one feel of content and messages throughout the app. Having an i18n file or something similar will help @saintsoldierx or others concerned with user experience to be able to easily edit and change user-facing messages. Also, it will help in having multilingual support for non-English speaking countries. I would love to see a French interface (for Canadian and french audience) or a Tamil interface for our southern friends. 🙏

inderpreetsingh:
- Separate all strings to separate place. 
- Makes the whole codebase compatible with separated strings (use new variables)

Fixes #565


www/index.html
```diff
@@ -203,5 +203,12 @@
     <script src="assets/js/vendor/mousetrap.min.js"></script>
     <script src="js/desktop_index.js"></script>
     <script type="text/javascript" src="js/vendor/tingle.js"></script>
+    <script>
+      const { remote } = require('electron');
+      const { i18n } = remote.require('./app');
+      document.querySelectorAll('[data-key]').forEach(el => {
```
tsingh777, line 209: 👍

www/js/desktop_scripts.js
```diff
@@ -167,7 +167,7 @@ module.exports = {
                             fs.unlink(oldDBPath, err1 => {
                               if (err1) {
                                 // eslint-disable-next-line no-console
-                                console.log(`Could not delete old database ${oldDB}: ${err1}`);
+                                console.log(i18n.t('DATABASE.COULDNT_DELETE', { oldDB, err1 }));
```
tsingh777, line 170: this does not need to translated.

www/index.html
```diff
@@ -203,5 +203,12 @@
     <script src="assets/js/vendor/mousetrap.min.js"></script>
     <script src="js/desktop_index.js"></script>
     <script type="text/javascript" src="js/vendor/tingle.js"></script>
+    <script>
+      const { remote } = require('electron');
+      const { i18n } = remote.require('./app');
```
tsingh777, line 208: this technically a bad practice. future versions of electron will remove this. its fine for now.

www/js/chromecast.js
```diff
@@ -128,7 +132,7 @@ function receiverMessage(namespace, message) {
  * session listener during initialization
  */
 function sessionListener(e) {
-  appendMessage('New session ID:' + e.sessionId);
+  appendMessage(i18n.t('CHROMECAST.NEW_SESSION_ID') + e.sessionId);
```
tsingh777, line 135: anything going to appendMessage doesn't need to be translated. console logs don't need to be translated.

www/js/markdownToHTML.js
```diff
@@ -16,4 +19,7 @@ function markdownToHTML(file) {
   }
 }
 
+const key = document.querySelector('h1').innerHTML;
```
tsingh777, line 22: ideally we should use the same data-keys function for all the html files.

tsingh777:
@maneetpaul this approach looks good. Can you help get another language for all the strings. https://github.com/inderpreetsingh/sttm-desktop/blob/i18n/www/locales/en/translation.json

inderpreetsingh:
> @maneetpaul this approach looks good. Can you help get another language for all the strings. https://github.com/inderpreetsingh/sttm-desktop/blob/i18n/www/locales/en/translation.json

@tsingh777 veerji - I think the motive of this PR was to separate strings in a separate file. We did it via i18n so that we can do translations in future. However for now it was meant to be a code refactor practice. We can have a switch option in next milestone or something. I say that only because tasks like these usually take a long time from non-dev side (gathering translations and all) and I would like us to be moving to next milestone.

maneetpaul:
> > @maneetpaul this approach looks good. Can you help get another language for all the strings. https://github.com/inderpreetsingh/sttm-desktop/blob/i18n/www/locales/en/translation.json
> 
> @tsingh777 veerji - I think the motive of this PR was to separate strings in a separate file. We did it via i18n so that we can do translations in future. However for now it was meant to be a code refactor practice. We can have a switch option in next milestone or something. I say that only because tasks like these usually take a long time from non-dev side (gathering translations and all) and I would like us to be moving to next milestone.

Agreed. We will do a translation in a future milestone.

app.js
```diff
@@ -22,6 +24,16 @@ const http = require('http-shutdown')(httpBase);
 const io = require('socket.io')(http);
 /* eslint-enable */
 
+// Configuring the i18n
+i18n.use(i18nBackend);
+i18n.init({
+  backend: {
+    loadPath: './www/locales/{{lng}}/{{ns}}.json',
```
tsingh777, line 31: can we swap the file name and the folder name? make the file `en.json` without the extra folder?

tsingh777:
> > @maneetpaul this approach looks good. Can you help get another language for all the strings. https://github.com/inderpreetsingh/sttm-desktop/blob/i18n/www/locales/en/translation.json
> 
> @tsingh777 veerji - I think the motive of this PR was to separate strings in a separate file. We did it via i18n so that we can do translations in future. However for now it was meant to be a code refactor practice. We can have a switch option in next milestone or something. I say that only because tasks like these usually take a long time from non-dev side (gathering translations and all) and I would like us to be moving to next milestone.

That's fine, I was asking if you tested programmatically changing the language. if we change it from the rendered process will it propagate to the main process and update other Browser windows as a person is using the app or will they have to change language and close then app for the changes take affect.

tsingh777:
> > > @maneetpaul this approach looks good. Can you help get another language for all the strings. https://github.com/inderpreetsingh/sttm-desktop/blob/i18n/www/locales/en/translation.json
> > 
> > 
> > @tsingh777 veerji - I think the motive of this PR was to separate strings in a separate file. We did it via i18n so that we can do translations in future. However for now it was meant to be a code refactor practice. We can have a switch option in next milestone or something. I say that only because tasks like these usually take a long time from non-dev side (gathering translations and all) and I would like us to be moving to next milestone.
> 
> Agreed. We will do a translation in a future milestone.

We should still start looking into starting the translation process.

inderpreetsingh:
> That's fine, I more concerned about if you tested programmatically changing the language. if we change it from the rendered process will it propagate to the main process and update other Browser windows as a person is using the app or will they have to change language and close then app for the changes take affect. We will need another language file (I guess gibberish would work too as a test).

@tsingh777 veerji - no i didn't test. My goal was just to separate strings in a different file, wasn't even thinking about languages yet.

PR #1061, opened 2020-04-13
adds a flag to prevent code changes

### LINKED ISSUE #1021: Controller code keeps changing
Multiple reports of the code being changed. Specifically with bad internet

inderpreetsingh:
Tries to fix #1021

Needs extensive testing.

### DIFF
--- a/www/js/toolbar.js
+++ b/www/js/toolbar.js
@@ -102,8 +102,8 @@ const remoteSyncInit = async () => {
     return;
   }
   if (onlineVal) {
-    isRequestSent = true;
     const newCode = await tryConnection();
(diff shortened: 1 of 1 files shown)


www/js/toolbar.js
```diff
@@ -97,7 +98,11 @@ const setListeners = () => {
 
 const remoteSyncInit = async () => {
   const onlineVal = await isOnline();
+  if (isRequestSent) {
+    return;
+  }
   if (onlineVal) {
+    isRequestSent = true;
```
tsingh777, line 105: maybe this should be after the next line

PR #1062, opened 2020-04-14
Anand sahib sync

inderpreetsingh:
Desktop side code of anand sahib sync. 

It needs some fixes on API side.

**Edit** Fixes Made

https://github.com/KhalisFoundation/sttm-web/pull/1033

Watch it at https://youtu.be/F1-oV3SIWMc

### DIFF
--- a/www/js/search.js
+++ b/www/js/search.js
@@ -948,6 +948,7 @@ module.exports = {
           englishHeading ? 'roman' : 'gurmukhi'
         }${seenClasses}`,
         {
+          'data-cp-id': line.crossPlatformID,
           'data-line-id': line.ID,
           'data-main-letters': line.MainLetters,
           onclick: e => {
--- a/www/js/search.js
+++ b/www/js/search.js
@@ -759,11 +759,12 @@ module.exports = {
     }
   },
 
-  async loadCeremony(ceremonyID, LineID = null, historyReload = false) {
+  async loadCeremony(ceremonyID, LineID = null, historyReload = false, crossPlatformID = null) {
     const $shabadList = this.$shabad || document.getElementById('shabad');
     $shabadList.innerHTML = '';
     $shabadList.dataset.bani = '';
     try {
+      let currentRow;
       const rowsDb = await banidb.loadCeremony(ceremonyID);
       const rows = await Promise.all(
         rowsDb.map(rowDb => {
@@ -786,6 +787,10 @@ module.exports = {
             row = banidb.loadVerses(rowDb.VerseIDRangeStart, rowDb.VerseIDRangeEnd);
           }
           row.sessionKey = `ceremony-${ceremonyID}`;
+          row.crossPlatformID = rowDb.ID;
+          if (row.crossPlatformID == crossPlatformID || row.ID == LineID) {
+            currentRow = row;
+          }
           return row;
         }),
       );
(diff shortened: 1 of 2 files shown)


www/js/search.js
```diff
@@ -931,7 +944,9 @@ module.exports = {
         'data-line-count': line.lineCount,
       },
       h(
-        `a#line${line.ID}.panktee.${englishHeading ? 'roman' : 'gurmukhi'}${seenClasses}`,
+        `a#line${line.ID}.cpid-${line.crossPlatformID}.panktee.${
```
tsingh777, line 947: why is cpid-${line.crossPlatformID} being added as a class and not as a `data-attribute`?

www/js/search.js
```diff
@@ -786,6 +789,13 @@ module.exports = {
             row = banidb.loadVerses(rowDb.VerseIDRangeStart, rowDb.VerseIDRangeEnd);
           }
           row.sessionKey = `ceremony-${ceremonyID}`;
+          row.crossPlatformID = rowDb.ID;
+          if (row.crossPlatformID === crossPlatformID || row.ID === LineID) {
```
tsingh777, line 793: the logic in this map function is a bit confusing. 

you getting all the verses for the ceremony, then adjusting the lineIDs based on the crossplatformID values and if the current line is being shown? 

can you add some comments?

www/js/search.js
```diff
@@ -1111,25 +1126,24 @@ module.exports = {
 
   clickShabad(e, ShabadID, LineID, Line, rows, mode = 'click') {
     if (window.socket !== undefined) {
-      let shabadIdsplit = [ShabadID];
-      if (typeof ShabadID === 'string') {
-        shabadIdsplit = ShabadID.split('-');
+      let sessionKeySplit;
```
tsingh777, line 1129: what is this used for?

inderpreetsingh: In this PR I changed name from shabadIdsplit to sessionKeySplit because that made more sense. As ceremonyID or baniID is not shabadID. 

However, in general a session Key is the ShabadID + session type. So for Ceremony it would be ceremony-ceremonyID, for bani it would be bani-baniLength-baniID.I have already  marked this code for improvement. Need better / simpler way to differentiate between 

a) Different type of shabads / verses (One routine to generally show shabad, and then some inherited routines to show ceremony, shabad, bani etc separately)
b) Different state of when they are loaded (search / from history etc). (Probably using one state management object rather than using bunch of global state variables) 

Currently all that is done by passing arguments which makes the whole logic little complex.

inderpreetsingh: Created a ticket for this improvement.

www/js/search.js
```diff
@@ -931,8 +952,11 @@ module.exports = {
         'data-line-count': line.lineCount,
       },
       h(
-        `a#line${line.ID}.panktee.${englishHeading ? 'roman' : 'gurmukhi'}${seenClasses}`,
+        `a#line${line.ID}.cpid-${line.crossPlatformID}.panktee.${
```
tsingh777, line 955: do you still need crossPlatformID as a `class`?

www/js/toolbar.js
```diff
@@ -69,6 +69,16 @@ const setListeners = () => {
         }
       };
 
+      const loadCeremony = (ceremonyId, crossPlatformId) => {
+        const currentCeremonyID = global.core.search.getCurrentShabadId();
+        const currentVerse = document.querySelector(`[data-cp-id = "${crossPlatformId}"]`);
```
tsingh777, line 74: 👍

PR #1073, opened 2020-05-06
Two sync

inderpreetsingh:
This allows a two way sync between desktop and web.

_video and more details coming soon_


www/js/search.js
```diff
@@ -22,7 +22,7 @@ const allowedKeys = [
 const sessionList = [];
 const sessionStatesList = {};
 const currentShabad = [];
-let currentShabadId = 0;
+let currentShabadState = 0;
```
tsingh777, line 25: maybe initialize this `{id: null, type: null}`? 

that way you won't get any "can't not access prop of" issues

www/js/toolbar.js
```diff
@@ -94,6 +94,34 @@ const setListeners = () => {
             type: 'response-control',
             success: isPinCorrect,
           });
+
+          // if Pin is correct and there is a shabad already in desktop, emit that shabad details.
+          if (isPinCorrect) {
+            const currentShabad = global.core.search.getCurrentShabadId();
+            const currentVerse = document.querySelector(`#shabad .panktee.current`);
+            const homeVerse = document.querySelector(`#shabad .panktee.main`);
```
tsingh777, line 102: I'm guessing updating the home line from desktop is out of scope right now?

maneetpaul: Not too worried about that if we can't add it right now

PR #1092, opened 2020-05-18
Sunder gutka sync support

inderpreetsingh:
Ability to work sunder gutka through sync / Bani Controller.

### DIFF
--- a/www/js/toolbar.js
+++ b/www/js/toolbar.js
@@ -80,6 +80,16 @@ const setListeners = () => {
         }
       };
 
+      const loadBani = (BaniId, crossPlatformId) => {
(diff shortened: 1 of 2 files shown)


www/js/toolbar.js
```diff
@@ -128,9 +140,7 @@ const setListeners = () => {
             isPinCorrect ? 'Connection Succesful' : 'Connection Failed',
           );
         },
-        /* Coming soon
-        'bani' : global.core.search.loadBani(data.baniId, data.verseId);
-        */
+        bani: payload => loadBani(payload.baniId, payload.verseId),
```
tsingh777, line 143: I need to spend some draw this architecture out. I'm losing track of how everything is loaded 😓

PR #1097, opened 2020-05-23
Add babel and partial react. 

tsingh777:


tsingh777:
Tested locally packaged version. looked ok.

PR #1109, opened 2020-06-03
adds green screen option to bani overlay

### LINKED ISSUE #1088: Create a green screen option for Bani Overlay
If a user is using Bani Overlay as a video input, they won't be able to use it as an overlay unless there is a chroma key channel. So make the white area (where video would appear) as green. 

Issue may be that if user selects 0.5 transparency (for example) for the actual overlaid banner/text, then it will mix in the green. If there is a way around it (so that the green chroma isn't overlaid) then great - else disable transparency settings and make it 1.0 (100%). If user wishes to make it transparent, i am sure their software they are using for mixing video will allow for transparency settings.

inderpreetsingh:
Fixes #1088 

https://youtu.be/eo8HeExaVwI

Adds a toggle for the Green Screen. When green screen is on, we switch off the opacity options.

### DIFF
--- a/www/js/defaults.json
+++ b/www/js/defaults.json
@@ -107,7 +107,8 @@
         "translationFont": "Muli",
         "gurbaniFont": "Gurbani Akhar Thick",
         "larivaarAssist": false,
-        "theme": false
+        "theme": false,
+        "greenScreen": false
       },
       "gurbani": {
         "bottom": 156,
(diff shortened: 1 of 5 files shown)


www/js/defaults.json
```diff
@@ -107,7 +107,8 @@
         "translationFont": "Muli",
         "gurbaniFont": "Gurbani Akhar Thick",
         "larivaarAssist": false,
-        "theme": false
+        "theme": false,
+        "greenScreen": false
```
tsingh777, line 111: make sure to check "fresh installs" we usually run into issue when adding new defaults.

PR #1132, opened 2020-06-16
Pagination problem refactor + Ceremonies Support

inderpreetsingh:
Change the quickThroughput to 20 and clean up some code. 
Also adds support for Long ceremonies (in future)

Can be used in place of the PR #1118


www/js/toolbar.js
```diff
@@ -52,9 +52,22 @@ const toggleOverlayUI = (toolbarItem, show) => {
...
     window.socket.on('data', data => {
       const isPinCorrect = parseInt(data.pin, 10) === adminPin;
 
+      const loadVerse = (crossPlatformId, lineCount) => {
+        $shabad.parentElement.scrollTo(0, parseInt(lineCount - 1, 10) * 35.6);
+        const currentVerse = document.querySelector(`[data-cp-id = "${crossPlatformId}"]`);
+        if (currentVerse) {
+          currentVerse.click();
+        } else {
+          store.set('GlobalState', {
```
tsingh777, line 65: ?. what are we adding to the store?

www/js/toolbar.js
```diff
@@ -52,9 +52,22 @@ const toggleOverlayUI = (toolbarItem, show) => {
 
 const setListeners = () => {
   if (window.socket !== undefined) {
+    const $shabad = document.getElementById('shabad');
     window.socket.on('data', data => {
       const isPinCorrect = parseInt(data.pin, 10) === adminPin;
 
+      const loadVerse = (crossPlatformId, lineCount) => {
+        $shabad.parentElement.scrollTo(0, parseInt(lineCount - 1, 10) * 35.6);
```
tsingh777, line 60: magic number 🌟

tsingh777: could use the variable you have `lineHeight`

www/js/toolbar.js
inderpreetsingh: We are adding the current verse selector which is independent of if you are in bani, ceremonies or Shabad, also independent of if you are accessing it via cross-platform ID.

tsingh777: I have hesitations with adding this to the store. if the use relaunches the app it would still be set and might not work as expected. should we clear it on app launch?

inderpreetsingh: Yes, just want to have a clean way to store a global state.

PR #1141, opened 2020-06-21
adds qr code functionality to bani controller

inderpreetsingh:
Screenshot 2020-06-21 at 6 59 51 PM


package.json
```diff
@@ -88,6 +88,7 @@
     "moment": "^2.22.1",
     "noty": "^3.2.0-beta",
     "openport": "0.0.5",
+    "qrcode": "^1.4.4",
```
tsingh777, line 91: missing package-lock.json change.

www/js/toolbar.js
```diff
@@ -327,6 +348,11 @@ const syncContent = h('div.sync-content', [
       false,
     ),
   ]),
+  h('div.qr-container', [
```
tsingh777, line 351: maybe we can get a head start in to componentization and have this in its own file and import it?

inderpreetsingh: Umm, yes, that's part of the toolbar refactor, all this is going to separate files anyway.

package.json
```diff
@@ -99,6 +99,7 @@
...
```
tsingh777, line 102: would be cool if we could add an image in the center. I don't think this module does that unfortunately.

inderpreetsingh: @tsingh777 Yes that would be so cool. :) Theoretically, we can, it's a canvas rendering, so it can be modified. I have visiting cards with such modified QR codes, and on some basic phones, it won't pick up the info. It actually depends upon the contrast of the image. If I can get a UI, I can give it a shot.

PR #1142, opened 2020-06-22
Fix live feed files path selection

tsingh777:
Electron update caused api change for using dialog. 

https://github.com/electron/electron/blob/v7.2.3/docs/api/dialog.md#dialogshowopendialogsyncbrowserwindow-options


maneetpaul:
Approving / Merging so I can QA with the rest of the release

PR #1148, opened 2020-06-24
changes hukamnama api

inderpreetsingh:
Changes hukamanama API to v2. Shows the hukamnama shabad. :)

### DIFF
--- a/www/main/menu.js
+++ b/www/main/menu.js
@@ -172,7 +172,7 @@ const hukamnamaButton = h(
     'a.hukamnama-button',
     {
       onclick: () => {
-        getJSON('https://api.banidb.com/v2/hukamnamas ', (error, response) => {
+        getJSON('https://api.banidb.com/v2/hukamnamas/today ', (error, response) => {
           if (!error) {
             const hukamShabadID = parseInt(response.shabadIds[0], 10);
             analytics.trackEvent('display', 'hukamnama', hukamShabadID);
--- a/www/main/menu.js
+++ b/www/main/menu.js
@@ -172,10 +172,9 @@ const hukamnamaButton = h(
     'a.hukamnama-button',
     {
       onclick: () => {
-        getJSON('https://api.banidb.com/hukamnama/today', (error, response) => {
+        getJSON('https://api.banidb.com/v2/hukamnamas ', (error, response) => {
           if (!error) {
-            const hukamShabadID = parseInt(response.shabadinfo.id, 10);
-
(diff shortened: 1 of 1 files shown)


www/main/menu.js
```diff
@@ -172,10 +172,9 @@ const hukamnamaButton = h(
     'a.hukamnama-button',
     {
       onclick: () => {
-        getJSON('https://api.banidb.com/hukamnama/today', (error, response) => {
+        getJSON('https://api.banidb.com/v2/hukamnamas/today ', (error, response) => {
```
tsingh777, line 175: /today does not add anything. the api is just giving you the latest. I'd leave it out and just call ```https://api.banidb.com/v2/hukamnamas```

tsingh777, line 175: We should only call this if the user has an internet connection. options are to 

- either hide the button if they are offline or <- my preference. 
- show some message that they must be online, or 
- pick a random shabad.

inderpreetsingh: I don't prefer hiding because that might result in the user trying to find "where did the hukamnama feature go?" But I will defer that decision to @maneetpaul veerji.

tsingh777: another would be to make it "disabled" if user is offline. we'll need to do the same for the notifications bell.

tsingh777: I don't agree with that. its clear from readability yes, but it gives false pretense on what the API provides. until the API changes we shouldn't be calling it in a way that it's not intended, even if it's the result we want. 

I'd prefer if we leave out "today" since the actual API uses what after the path and expect it to be a date.

tarunsingh5: @ManjotS do you know when we will have https://api.banidb.com/v2/hukamnamas/{date} functionality?

ManjotS:
We have had that for years and its what drives the yesterday/tomorrow
buttons on the hukamnama sttm web

https://api.banidb.com/v2/hukamnamas/2020/07/15

www/main/menu.js
tarunsingh5: from @ManjotS "We have had that for years and its what drives the yesterday/tomorrow
buttons on the hukamnama sttm web

https://api.banidb.com/v2/hukamnamas/2020/07/15"

tarunsingh5: @ManjotS /yesterday and /today give the same result though... a bug?

ManjotS:
There is no yesterday. It's either by date, or today.

tsingh777:
fixed in https://github.com/KhalisFoundation/sttm-desktop/pull/1268

PR #1162, opened 2020-06-30
Adds a spinner in sync modal

inderpreetsingh:
- show loader when waiting for code
- if code returns false, then restart the remote sync

https://youtu.be/S4X_9HkdZAY

### DIFF
--- a/www/locales/en.json
+++ b/www/locales/en.json
@@ -317,6 +317,7 @@
     },
     "SYNC_CONTROLLER": {
       "INTERNET_ERR": "Sorry! You seem to be offline. Sync only works when you are connected to the internet.",
+      "CODE_ERR": "Sorry! Failed to get a new code for you, please try again later.",
       "PIN": "Pin",
       "PRESENT_CODE": "Present Code to Sangat",
       "SANGAT_SYNC": "Sangat Sync",
--- a/www/locales/en.json
+++ b/www/locales/en.json
@@ -316,8 +316,8 @@
       "2": " from a mobile device to search, navigate, and control the entire app."
     },
     "SYNC_CONTROLLER": {
-      "INTERNET_ERR": "Sorry! You seem to be offline. Sync only works when you are connected to the internet.",
-      "CODE_ERR": "Sorry! Failed to get a new code for you, please try again later.",
+      "INTERNET_ERR": "Sorry, You seem to be offline. Sync only works when you are connected to the internet.",
+      "CODE_ERR": "Sorry, Failed to get a new code for you, please try again later.",
       "PIN": "Pin",
       "PRESENT_CODE": "Present Code to Sangat",
       "SANGAT_SYNC": "Sangat Sync",
--- a/www/main/share-sync.js
+++ b/www/main/share-sync.js
@@ -1,5 +1,9 @@
 const request = require('request-promise');
-const { store } = require('electron').remote.require('./app');
+const { remote } = require('electron');
+const { store } = remote.require('./app');
+const analytics = remote.getGlobal('analytics');
+
+const Noty = require('noty');
 
 const SYNC_API_URL = 'https://api.sikhitothemax.org';
 const SOCKET_SCRIPT_SOURCE = `${SYNC_API_URL}/socket.io/socket.io.js`;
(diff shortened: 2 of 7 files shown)


www/main/toolbar.js
```diff
@@ -281,51 +288,54 @@ const adminContent = h('div', [
...
-    h(
-      'button.button.copy-code-btn',
-      {
-        onclick: () => {
-          if (code !== '...') {
-            const syncString = i18n.t('TOOLBAR.SYNC_CONTROLLER.SYNC_STRING', { code });
-            global.controller.sendText(syncString);
-            analytics.trackEvent('controller', 'codePresented', true);
-          }
+const syncContent = h('div.sync-content-wrapper', [
```
tsingh777, line 291: would have been cool to do this as a react component.

www/main/toolbar.js
```diff
@@ -153,12 +153,19 @@ const setListeners = () => {
...
   const onlineVal = await isOnline();
   if (isRequestSent) {
     return;
   }
   if (onlineVal) {
     const newCode = await tryConnection();
+    if (newCode) {
+      $syncConent.classList.remove('loading');
+    } else {
+      remoteSyncInit();
```
tsingh777, line 167: this could get stuck in an endless loop if the service is down. We should handle the error better in tryConnection.

www/main/toolbar.js
```diff
@@ -281,51 +288,54 @@ const adminContent = h('div', [
...
-      () => {
-        syncToggle();
-        analytics.trackEvent('controller', 'connection', isConntected ? 'Enabled' : 'Disabled');
-      },
-      false,
+    syncItemFactory(
+      i18n.t('TOOLBAR.BANI_CONTROLLER'),
+      h('span', [
+        i18n.t('TOOLBAR.BANI_DESC.1'),
+        h('strong', 'sttm.co/control'),
```
tsingh777, line 321: I wonder if we should make this an anchor tag?

www/main/toolbar.js
```diff
@@ -281,51 +288,54 @@ const adminContent = h('div', [
...
+        i18n.t('TOOLBAR.BANI_DESC.2'),
+      ]),
+      adminContent,
     ),
+    h('div.connection-switch-container', [
+      h('p', i18n.t('TOOLBAR.DISABLE_CONNECTIONS_MSG', { appName: i18n.t('APPNAME') })),
```
tsingh777, line 327: i'm wondering if   "APPNAME": "SikhiToTheMax", needs to be translated.

www/main/toolbar.js
inderpreetsingh: Yes, but moving the whole thing to components would require somewhat bigger refactor that I am tracking in a separate ticket / branch. We wanted to do this in current release, so I just switched back to dev branch and fixed this first.

www/locales/en.json
```diff
@@ -317,6 +317,7 @@
     },
     "SYNC_CONTROLLER": {
       "INTERNET_ERR": "Sorry! You seem to be offline. Sync only works when you are connected to the internet.",
+      "CODE_ERR": "Sorry! Failed to get a new code for you, please try again later.",
```
tsingh777, line 320: I think there's a rule that we shouldn't use exclamation in text for the user.

www/main/share-sync.js
```diff
@@ -19,26 +24,44 @@ module.exports = {
...
-      const {
-        data: { namespaceString },
-      } = JSON.parse(result);
-
-      if (window.io !== undefined) {
-        window.namespaceString = namespaceString;
-        onConnect(namespaceString);
-      } else {
-        // TODO: Wait for io or something
+    const getNewCode = async () => {
```
tsingh777, line 29: does this function need to be in tryConnection?

www/main/share-sync.js
```diff
@@ -19,26 +24,44 @@ module.exports = {
...
+      return newCode;
+    };
+
+    // if a succesful code already exists, use that or else get new code
+    try {
+      await request(`${SYNC_API_URL}/sync/join/${window.namespaceString}`);
+      syncCode = window.namespaceString;
     } catch (e) {
-      return false;
+      syncCode = getNewCode();
```
tsingh777, line 61: do you need to try the join call again?

www/locales/en.json
inderpreetsingh: Oh, wasn't aware of that rule. Will change it. Where can I read these rules?

www/main/share-sync.js
inderpreetsingh: No, it doesn't need to be.

www/main/share-sync.js
inderpreetsingh: Yes, so when the existing code has stopped working we need to get a new code.

www/main/toolbar.js
inderpreetsingh: @tarunsingh5 yes, i think it would be transliterated.

www/main/toolbar.js
inderpreetsingh: Good idea, we can do that.

www/main/share-sync.js
```diff
@@ -17,28 +49,20 @@ module.exports = {
...
-        onConnect(namespaceString);
-      } else {
-        // TODO: Wait for io or something
-      }
-      return namespaceString;
+      await request(`${SYNC_API_URL}/sync/join/${window.namespaceString}`);
+      syncCode = window.namespaceString;
     } catch (e) {
-      return false;
+      syncCode = getNewCode(host);
```
tsingh777, line 62: sorry, didn't mean try as in a `try` block. I mean do you need to call `/sync/join/` after this function?

inderpreetsingh: Umm, no I don't think we need to do that, once we get the new code from API, it works, and it's checked every time user opens the sync dialogue in app. Adding another join here would be helpful in case API returns a code but its instantly not usable. Which sounds like would be a bigger problem on API side than what we need to handle. What do you suggest? Am I missing something?

ManjotS:
I was talking to tsingh about this.. maybe api should add a resume
function? Basically if you hit a socket and its not there, instead of doing
begin again, do a resume. This would allow us to not annoy the user.

>

www/main/share-sync.js
tsingh777: /begin creates the socket and returns the namespace. /join is how the app connects to the socket. The try block is doing /join so I figured when we get to the catch, we want to recreate the socket and join it?

tsingh777:
> I was talking to tsingh about this.. maybe api should add a resume function? Basically if you hit a socket and its not there, instead of doing begin again, do a resume. This would allow us to not annoy the user.
> […](#)

that might require a refactor. we kill sockets after all clients disconnect.

PR #1178, opened 2020-08-03
fixes #1165 adds new theme nirbaan

### LINKED ISSUE #1165: Add new theme
"Nirbaan"

nirbaan-bg
image
color #s - ffebbe c2af85 9cabff ff723b
visraams - ffeb3b ff723b

inderpreetsingh:
Screenshot 2020-08-03 at 7 08 32 PM


tsingh777:
Need to add this to chromecast and overlay as well.

www/main/overlay.js
```diff
@@ -7,7 +7,7 @@ const { obs: defaultPrefs } = require('./configs/defaults.json');
 
 const { ipcRenderer, remote } = electron;
 
-const { store, i18n } = remote.require('./app');
+const { store, i18n, overlayPresets } = remote.require('./app');
```
tsingh777, line 10: does this need to come through `remote.require('./app')`? can't you include the json file directly here since it's only used in this window?

inderpreetsingh: yeah you are right!

inderpreetsingh: Fixed

PR #1179, opened 2020-08-03
Settings for sync

inderpreetsingh:
Setting up settings object
* When connected
* When recieving trigger to change settings from web.

General API format for settings.

```
action: `changeFontSize`
target: `translation`
value: `increase`
```

### DIFF
--- a/www/main/toolbar.js
+++ b/www/main/toolbar.js
@@ -131,6 +131,9 @@ const setListeners = () => {
             host: 'sttm-desktop',
             type: 'response-control',
             success: isPinCorrect,
+            settings: {
+              fontSizes: store.getUserPref('slide-layout.font-sizes'),
+            },
           });
 
           // if Pin is correct and there is a shabad already in desktop, emit that shabad details.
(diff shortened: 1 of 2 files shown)


www/main/toolbar.js
```diff
@@ -170,6 +173,12 @@ const setListeners = () => {
         },
         bani: payload => loadBani(payload.baniId, payload.verseId, payload.lineCount),
         ceremony: payload => loadCeremony(payload.ceremonyId, payload.verseId, payload.lineCount),
+        settings: payload => {
+          const { settings } = payload;
+          if (settings.action === 'changeFontSize') {
+            global.core.menu.settings.changeFontSize(settings.target, settings.value);
```
tsingh777, line 179: can you comment on what the payload looks like when it sends settings? Does documentation need to be updated?

inderpreetsingh: yes documentation needs to be updated.

inderpreetsingh:
Sorry @tsingh777 veerji, I will have to make a change. Just found a bug in testing.

PR #1181, opened 2020-08-12
Add Realm search for Main letters

tsingh777:
This adds the the realm query needed for Main letter search. It does not expose it.


inderpreetsingh (approved):
LGTM 💯

PR #1202, opened 2020-09-09
Exposes the zoom overlay setting in overlay

Gauravjeetsingh:
image

Few concerns/next steps which we can either do in this PR or have separate PRs for them.

 - Do we want to remove the Zoom Settings page which @tsingh777 veerji originally created? it has a little more detailed explanation about setting up captions with links, so I didn't change anything in them yet.

Deleted the previous zoom settings page

- Should there be an indicator when the sttm-desktop is connected to a Zoom meeting?

- Do we need a separate button to clear the zoom API key? It was in the original zoom cc page but wasn't in the design Aman veerji shared, and even I think we don't need a separate button to clear it. Let me know if it had a particular use case.

Added the clear button, when we save.

- Where should the red Instructions button point to? Would that be an external link or it will be a page in the sttm-desktop itself.

It would be an external link, not added yet.


app.js
```diff
@@ -65,6 +66,8 @@ if (currentTheme === undefined) {
   store.setUserPref('app.theme', themes[0].key);
 }
 
+store.set('userPrefs.app.zoomToken', '');
```
tsingh777, line 69: lets move this to when the app is ready so we don't run into that issue with "new" installs that don't have the preferences files.

app.js
```diff
@@ -129,6 +133,7 @@ function openSecondaryWindow(windowName) {
       if (window.focus) {
         window.focus();
       }
+      window.obj.webContents.openDevTools();
```
tsingh777, line 136: remove

app.js
```diff
@@ -112,6 +115,7 @@ function openSecondaryWindow(windowName) {
     openWindow[0].show();
   } else {
     window.obj = new BrowserWindow({
+      minWidth: 800,
```
tsingh777, line 118: think we can remove this too.

app.js
```diff
@@ -343,6 +350,24 @@ const showLine = (line, socket = io) => {
   if (!lineWithSettings.fromScroll) {
     socket.emit('show-line', payload);
   }
+  const zoomToken = store.get('userPrefs.app.zoomToken');
+  const english =
+    store.get('userPrefs.slide-layout.fields.display-translation') &&
+    store.get('userPrefs.slide-layout.language-settings.translation-language');
+  if (zoomToken) {
+    try {
+      // console.log(`Sending-${line.Line.Unicode}\n${english ? line.Line.English : ''}`);
```
tsingh777, line 359: remove

app.js
```diff
@@ -343,6 +350,24 @@ const showLine = (line, socket = io) => {
...
+      await fetch(`${zoomToken}&seq=${seq}`, {
+        method: 'POST',
+        body: `${line.Line.Unicode}\n${english ? line.Line.English : ''}\n`,
+      });
+      seq += 1;
+    } catch (e) {
+      // console.log(e);
```
tsingh777, line 366: remove this line.

www/locales/en.json
```diff
@@ -304,6 +305,7 @@
     "INSERT": "Insert",
     "LOCKED_SCREEN": "SikhiToTheMax is being controlled remotely.",
     "MOBILE_DEVICE_SYNC": "Mobile Device Sync",
+    "ZOOM_HEADING": "<img class='zoom-logo' src='assets/img/icons/zoom.svg'> Overlay",
```
tsingh777, line 308: this is image tag shouldn't in the language file. we use it inline?

tsingh777 (changes requested):

www/main/toolbar.js
```diff
@@ -323,6 +323,57 @@ const adminContent = h('div', [
...
+        onclick: () => {
+          // clears the zoom api token
+          document.querySelector('.zoom-api-input').value = '';
+          store.set('userPrefs.app.zoomToken', '');
+          document.querySelector('.save-btn').classList.remove('hidden-btn');
+          document.querySelector('.clear-btn').classList.add('hidden-btn');
+        },
+      }),
+    ]),
+    h('button.instructions-btn', [
```
tsingh777, line 359: potentially hide this until we have the a link to go to.

tsingh777 (approved):
LGTM

PR #1211, opened 2020-09-24
Centers the text in sundar gutka logo

Gauravjeetsingh:
Haven't tested in windows, but that's how it looks in Linux now.

image


maneetpaul:
@Gauravjeetsingh Veerjee the color looks different in this screenshot - seems more purple than blue.

PR #1234, opened 2020-11-03
Workspaces 😳✌🏼😍

inderpreetsingh:
Screenshot 2020-11-03 at 5 46 33 PM
Screenshot 2020-11-03 at 5 46 43 PM

### DIFF
--- a/www/main/workspace-bar/components/WorkspaceBar.jsx
+++ b/www/main/workspace-bar/components/WorkspaceBar.jsx
@@ -1,0 +1,40 @@
+import React, { useState } from 'react';
+import { remote } from 'electron';
+
+const { store } = remote.require('./app');
+
+const WorkspaceBar = () => {
+  const workspaces = ['Presenter', 'Single Screen'];
+  const defaultWsState = store.getUserPref('app.layout.presenter-view')
+    ? workspaces[0]
+    : workspaces[1];
+  const [currentWorkspace, setWorkspace] = useState(defaultWsState);
+
+  const handleWorkspaceChange = workspace => {
+    const moveToPresenter = workspace === 'Presenter';
+    store.setUserPref('app.layout.presenter-view', moveToPresenter);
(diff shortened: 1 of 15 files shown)


www/main/workspace-bar/components/WorkspaceBar.jsx
```diff
@@ -0,0 +1,40 @@
+import React, { useState } from 'react';
+import { remote } from 'electron';
+
+const { store } = remote.require('./app');
+
+const WorkspaceBar = () => {
+  const workspaces = ['Presenter', 'Single Screen'];
+  const defaultWsState = store.getUserPref('app.layout.presenter-view')
+    ? workspaces[0]
+    : workspaces[1];
```
tsingh777, line 10: we should make this more scalability.  `app.layout.presenter-view` won't work when we have more tabs. maybe we can transition to a new UserPrefKey?

www/main/workspace-bar/components/WorkspaceBar.jsx
```diff
@@ -0,0 +1,40 @@
+import React, { useState } from 'react';
+import { remote } from 'electron';
+
+const { store } = remote.require('./app');
+
+const WorkspaceBar = () => {
+  const workspaces = ['Presenter', 'Single Screen'];
```
tsingh777, line 7: Should be translated.

maneetpaul: Can we also change "Presenter" to "Presentation", and "Single Screen" to "Single Display"

www/main/workspace-bar/components/WorkspaceBar.jsx
inderpreetsingh: We are changing the settings infrastructure, so will change this along with that PR.

tsingh777 (approved):

PR #1247, opened 2020-11-19
Minimizes the workspace bar on minimizing navigator

### LINKED ISSUE #1240: Workspace bar and Toolbar should minimize
When in Single Display and pressing the minimize button, both the toolbar and workspace bar should hide away until the search pane is brought up again

Gauravjeetsingh:
Peek 2020-11-19 19-57

Fixes #1240

### DIFF
--- a/www/src/scss/_workspace-bar.scss
+++ b/www/src/scss/_workspace-bar.scss
@@ -7,6 +7,7 @@
   left: 0;
   line-height: 30px;
   position: fixed;
+  transition: height 600ms cubic-bezier(0.165, 0.84, 0.44, 1);
   top: 0;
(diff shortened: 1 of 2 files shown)


maneetpaul:
In addition to this, we discussed having the toolbar stay fixed to the left side of the screen across all workspaces. And pressing the minimize button would hide both the toolbar and workspace bar.

maneetpaul:
> In addition to this, we discussed having the toolbar stay fixed to the left side of the screen across all workspaces. And pressing the minimize button would hide both the toolbar and workspace bar.

This will be a sep PR

tsingh777 (approved):

Issue #1241, opened 2020-11-12
Bug: Toolbar doesn't come back when switching workspaces


maneetpaul:
If in Single Display workspace with the search pane minimized, and then switching to Presentation, the toolbar should come back.

PR #1248, opened 2020-11-20
Toolbar Changes

### LINKED ISSUE #1239: Toolbar should be the same in both workspaces
Single Display workspace shouldn't have a "mini" toolbar. Keep it fixed to the left side like in Presentaiton

inderpreetsingh:
* change position of toolbar in single display mode (make it like presentation mode) Fixes #1239 
* Solved Quicktool bugs that arose because of the above change
* Adds presenter view flag now in the viewer as well
* Make sure toolbar doesn't hide in presenter view #1241

### DIFF
--- a/www/main/desktop_scripts.js
+++ b/www/main/desktop_scripts.js
@@ -231,6 +231,8 @@ if ($minimize) {
         element.classList.toggle('disabled');
       });
       document.getElementById('navigator').classList.toggle('minimized');
+      if (global.webview) global.webview.send('navigator-toggled');
+      if (global.platform) global.platform.ipc.send('navigator-toggled');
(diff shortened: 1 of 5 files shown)


www/main/desktop_scripts.js
```diff
@@ -231,6 +231,7 @@ if ($minimize) {
         element.classList.toggle('disabled');
       });
       document.getElementById('navigator').classList.toggle('minimized');
+      if (global.webview) global.webview.send('navigator-toggled');
```
tsingh777, line 234: nit: prefer if blocks with curly brackets.

tsingh777 (approved):

PR #1252, opened 2020-12-07
Hides the quick tools in single window mode

Gauravjeetsingh:
Peek 2020-12-07 14-50


www/src/scss/viewer.scss
```diff
@@ -352,14 +352,21 @@ h2 {
 .quick-tools {
   .viewer-controls,
   &.webview .vc-toggle-icon {
-    display: flex;
+    display: none;
     left: 75px;
     opacity: 0.9;
```
inderpreetsingh, line 357: May be remove the whole thing in single display window rather than just changing the `display:none` ?

Also did you check changing the quick tool setting? Won't it re-appear?

inderpreetsingh (changes requested):

www/src/scss/viewer.scss
Gauravjeetsingh: Yes, it didn't reappear with setting. Also, with recent commit it now hides the quick tools setting as well. I will add the similar css hide thing in newer setting PR as well. Since it has different class for each setting.

inderpreetsingh (approved):
💯 LGTM, should QA

PR #1254, opened 2020-12-08
Adds the functionality to clear shabads from history

Gauravjeetsingh:
**In presenter view, we get a cross icon to remove specific shabads from the history plus one button to clear entire history.**
Peek 2020-12-08 21-08

**In single view, we dont have a single button to remove all history panktis, but we can remove them using the cross icon.**
Peek 2020-12-08 21-10

### DIFF
--- a/www/main/search.js
+++ b/www/main/search.js
@@ -634,6 +634,15 @@ module.exports = {
     }
   },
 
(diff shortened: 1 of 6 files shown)


inderpreetsingh (approved):
💯

PR #1268, opened 2021-01-04
Hukamnama changes to fix bug and better UX

inderpreetsingh:
-makes hukamnama button look disabled when offline
-changes api endpoint to v2

### DIFF
--- a/www/main/menu.js
+++ b/www/main/menu.js
@@ -4,8 +4,9 @@ const request = require('request');
 const moment = require('moment');
 const electron = require('electron');
 const sanitizeHtml = require('sanitize-html');
-const strings = require('./strings');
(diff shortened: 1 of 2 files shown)


www/main/menu.js
```diff
@@ -330,6 +335,10 @@ module.exports = {
     $listOfShabadOptions.appendChild(hukamnamaButton);
     $listOfShabadOptions.appendChild(notificationButton);
 
+    isOnline().then(online => {
```
tsingh777, line 338: Would the app need to be restarted to re-enabled the button if the user comes online? should be fine but just wondering if we should make that better.

tsingh777 (approved):

Issue #1293, opened 2021-02-10
Error Pop Up Coming up in 8.3


maneetpaul:
![Screen Shot 2021-02-10 at 11 41 48 AM](https://user-images.githubusercontent.com/10658257/107541443-23cf0100-6b95-11eb-8229-164b6e5e7777.png)

tsingh777:
This happens when closing changelog @inderpreetsingh

PR #1298, opened 2021-02-17
Analytics bug

### LINKED ISSUE #1294: Sundar Gutka and Ceremonies Not Loading
Might be because I am on an M1 Mac Mini. Need to investigate, could also be related to the refactor

inderpreetsingh:
Fixes #1294 #1293 

Still testing some things.

### DIFF
--- a/analytics.js
+++ b/analytics.js
@@ -31,7 +31,7 @@ class Analytics {
     const useragent = this.store.get('user-agent');
 
     if (process.env.NODE_ENV !== 'development') {
-      if (//global.getUserSettings.collectStatistics) {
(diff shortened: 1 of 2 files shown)


analytics.js
```diff
@@ -31,7 +31,7 @@ class Analytics {
     const useragent = this.store.get('user-agent');
 
     if (process.env.NODE_ENV !== 'development') {
-      if (global.getUserSettings.collectStatistics) {
+      if (this.store.get('userPrefs.app.analytics.collectStatistics')) {
```
tsingh777, line 34: this should be good for now. Let's think of how to deprecate the old store completely.

tsingh777 (approved):

PR #1300, opened 2021-02-18
removes write file from getSavedUserSettings

inderpreetsingh:
There was error on fresh install of 8.3 saying that user data did not exist. Fixed that error with this code change.

Earlier getSavedUserSettings, used to write the defaults to the file,if it didn't exist.. using writeFileSync, I realized that it's not needed because we are already writing to file in setUserSettingsState.

### DIFF
--- a/www/main/common/store/user-settings/get-saved-user-settings.js
+++ b/www/main/common/store/user-settings/get-saved-user-settings.js
@@ -20,7 +20,6 @@ function parseDataFile(filePath) {
     Object.keys(settings).forEach(key => {
       defaultSettings[key] = settings[key].initialValue;
     });
-    fs.writeFileSync(userConfigPath, JSON.stringify(defaultSettings));
     return defaultSettings;
   }
 }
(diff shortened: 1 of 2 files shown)


www/main/common/store/user-settings/get-saved-user-settings.js
```diff
@@ -20,7 +20,6 @@ function parseDataFile(filePath) {
     Object.keys(settings).forEach(key => {
       defaultSettings[key] = settings[key].initialValue;
     });
-    fs.writeFileSync(userConfigPath, JSON.stringify(defaultSettings));
```
tsingh777, line 23: so we won't write this file until a setting is changed? should be fine...?

tsingh777 (approved):

www/main/common/store/user-settings/get-saved-user-settings.js
inderpreetsingh: Yes, that's how the store used to work previously.

PR #1304, opened 2021-02-25
Fixes the live feed bug

Gauravjeetsingh:
Fixes the creation of the live feed files, from the new settings.

### DIFF
--- a/www/main/controller.js
+++ b/www/main/controller.js
@@ -582,7 +582,7 @@ module.exports = {
       mode,
       fromScroll,
     };
-    if (document.body.classList.contains('livefeed')) {
+    if (document.body.classList.contains('live-feed-true')) {
       showLinePayload.live = true;
     }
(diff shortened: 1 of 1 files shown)


www/main/controller.js
```diff
@@ -582,7 +582,7 @@ module.exports = {
       mode,
       fromScroll,
     };
-    if (document.body.classList.contains('livefeed')) {
+    if (document.body.classList.contains('live-feed-true')) {
```
tsingh777, line 585: could we not change this to check the setting instead of a class on the html?

Gauravjeetsingh: @tsingh777 @inderpreetsingh veerjis, updated this if condition. We only save the location in store, so I set its value to false when we turn off live feed. And used the location setting in this if condition.

tsingh777 (approved):

Issue #1382, opened 2021-11-19
9.0 Akhand Paath View Bug


MySmilingTurban:
It comes out as follows

![image](https://user-images.githubusercontent.com/51584037/142694586-41dee2c1-681e-4e09-afb5-9b82cf34b1c4.png)

Gauravjeetsingh:
fixed in dev branch, closing this out.

PR #1384, opened 2021-11-23
fix: Akhand path view display pane issue

theamanjs:
Refs: #1382

### DIFF
--- a/www/src/scss/viewer/shabad-deck/_shabad-deck.scss
+++ b/www/src/scss/viewer/shabad-deck/_shabad-deck.scss
@@ -3,9 +3,9 @@
     background-attachment: fixed;
     background-size: cover;
     height: unset;
+    overflow-y: scroll;
     padding-bottom: 10px;
     position: fixed;
-    overflow-y: scroll;
     width: 100%;
   }
 }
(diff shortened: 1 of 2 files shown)


www/src/scss/viewer/shabad-deck/_shabad-deck.scss
```diff
@@ -4,12 +4,16 @@
     background-size: cover;
     height: unset;
     padding-bottom: 10px;
+    position: fixed;
+    overflow-y: scroll;
```
Gauravjeetsingh, line 8: @theamanjs veerji, everything looks fine here. Just a minor change, can you switch up the order of CSS properties. overflow-y should be before position.

Gauravjeetsingh (changes requested):

www/src/scss/viewer/shabad-deck/_shabad-deck.scss
theamanjs: @Gauravjeetsingh thank you for your feedback veerji, I'm doing it right now :smile:

theamanjs: @Gauravjeetsingh veerji, I have made the changes.

Gauravjeetsingh (approved):

Gauravjeetsingh:
@theamanjs Instead of sending PRs to dev branch, can you make it to **version-9-qa-fixes** branch. We will merge to dev after all fixes are complete.

Issue #1362, opened 2021-10-15
9.0 Full results not visible in Search Results


MySmilingTurban:
Full results aren't visible in search results
The last few results are hidden beside other components (maybe footer)

**Steps to replicate**
Search 'qbAn' and number of results is written as 17 but if you count all of them, only 15 are visible

Happens when number of results are more than or equal to 5.

MySmilingTurban:
Waheguru jio,
Just a pinch of this one is left.
The last item of scroll views of search results and verses pane is partially visible.

inderpreetsingh:
> Waheguru jio, Just a pinch of this one is left. The last item of scroll views of search results and verses pane is partially visible.

Veerji, can you share the screenshot here?

MySmilingTurban:
![image](https://user-images.githubusercontent.com/51584037/140793825-c2d36f1a-811f-4104-848e-c72bfa05c40c.png)

MySmilingTurban:
Still only half part is visible of last element in the three scrollable panes (shown in pic)
Also, in the display window in presentation mode, there is white spacing appearing now

![image](https://user-images.githubusercontent.com/51584037/142693207-f841b375-d484-4ea7-8e19-a0e20b49d9aa.png)

inderpreetsingh:
Just a little more fine tuning, we can do this if this doesn't include any major change in code.

If the tools are opened at bottom of history, it still hides the last verse of history. You would need to subtract 180px from height if its opened, else 80px, if that's something we can do, go ahead, otherwise I think this works well. 

![Screenshot from 2021-12-01 01-28-41](https://user-images.githubusercontent.com/1131610/144119194-8ce13ede-9e0a-402f-b386-22b8a3ac7673.png)

Gauravjeetsingh:
Fixed in dev, closing this out

PR #1387, opened 2021-11-30
fix: hidden verse issue on windows

theamanjs:
Refs: #1362

### DIFF
--- a/www/main/navigator/misc/components/HistoryPane.jsx
+++ b/www/main/navigator/misc/components/HistoryPane.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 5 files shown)


inderpreetsingh:
Just a little more fine tuning, we can do this if this doesn't include any major change in code.

If the tools are opened at bottom of history, it still hides the last verse of history. You would need to subtract 180px from height if its opened, else 80px, if that's something we can do, go ahead, otherwise I think this works well. 

Screenshot from 2021-12-01 01-28-41

inderpreetsingh:
@theamanjs Can you fix the animation of shortcut tray?

theamanjs:
Yes veerji, I'm currently working on it. Sending the code in 30 mins.

Gauravjeetsingh (approved):

PR #1401, opened 2021-12-28
Fix the shabad viewer on first run and custom theme fix

inderpreetsingh:
This fixes the issue of showing the first Shabad that was broken after updating the realm code.


www/main/navigator/utils/load-shabad.js
```diff
@@ -1,5 +1,10 @@
 import banidb from '../../banidb';
 
 export const loadShabad = (shabadID, lineID) => {
-  return banidb.loadShabad(shabadID, lineID).then(rows => rows);
+  return banidb
+    .loadShabad(shabadID, lineID)
+    .then(rows => rows)
+    .catch(() => {
+      console.log('err while loading shabad, please contact dev team');
```
Gauravjeetsingh, line 8: @inderpreetsingh Do we need to show this to end user?

inderpreetsingh: we can do that in a separate PR, let's merge this fix so that its not a blocker in QA.

inderpreetsingh:
ezgif com-gif-maker

PR #1402, opened 2021-12-30
Revert to request module

Gauravjeetsingh:
With change in certificate, we added a manual certificate to be compatible with let's encrypt new root certificate.

But since it affected older versions as well, we changed the certificate. We don't need to check the certificate manually now. This PR reverts the code back to original request module.


inderpreetsingh (approved):
Looks good to me!

PR #1407, opened 2022-01-14
Fix viewer padding issue on windows single display mode

### LINKED ISSUE #1405: Padding issue in Single Display Mode on windows
Due to the additional title bar in windows, the content on the viewer is getting hidden.

Steps to reproduce:
- Go to single display mode
- Search any Shabad. Eg: `ddeh`
- You will see some part of the viewer is covered from the top.

Suggested solution:
- Add extra padding on top when it's on windows

image

theamanjs:
This PR fixes #1405 

The approach I used is by adding win32 class to the viewer when it's on windows and then adding padding-top on windows only. I am not sure if we need to do this on mac or not. Although It was written we need to do it on windows.

image


theamanjs:
Thank you @inderpreetsingh for helping out in this.

PR #1519, opened 2022-04-05
fix: shortcut tray state in history and tray component

theamanjs:
Fixes history bottom margin issue and links shortcut state with tray component


theamanjs:
Got some issues in PR

Issue #1570, opened 2022-05-06
History is not remembering what line user was on


saintsoldierx:
Users say they used to open up bani like Asa Di Vaar , then search for Shabads in between, and when they used to go back to Asa Di Vaar, it used to pick up from where they left it. It does not do that anymore. This could be related to history.

Gauravjeetsingh:
Closing this out, as its released to prod.

PR #1585, opened 2022-05-26
Fix the history issue

Gauravjeetsingh:
- Fixes the issue #1570 
- Reverses the order of history to newest first
- Fixes the home verse

Peek 2022-05-26 16-31


theamanjs (commented):
@Gauravjeetsingh veerij, everything looks good, though I'm seeing an issue when we open a shabad from history it actually creates the last read verse as the home verse. The rest of the things are looking pretty good.

inderpreetsingh (approved):

theamanjs (approved):
Looks good to me veerji

PR #1603, opened 2022-06-30
Fixes QA issues 1598, 1599, 1600

### LINKED ISSUE #1598: Preview in settings doesn't show Vishraam when Larivaar is disabled
**OS: Windows 10 & macOS**

This issue has similarities with #1527 for the preview window specifically.

With Larivaar and Larivaar assist disabled
image

With Larivaar disabled but Larivaar assist enabled
image

Steps to reproduce:
1. Search "KmrljA" and choose first shabad (the specific shabad was chosen due to easy visibility of both window highlights)
2. Enable Vishraams, both preview and display window shows the desired output
3. Now, enable larivaar to enable larivaar assist then disable larivaar
4. The preview window in settings shows no highlights (bug) whereas the display window shows the desired output with the vishraams.

### LINKED ISSUE #1599: Different font size of placeholder text in Gurmukhi and English
image

Steps to reproduce:
1. Open the app, clear search to see placeholder text
2. Switch from Gurmukhi to English and back, notice the little but weird font size difference.

### LINKED ISSUE #1600: {Windows} Scrollablility issue while using Filters
**OS: Windows 10**

In the following screenshots, the last results aren't visible while using filters:
image
image

Steps to reproduce:
1. Search "hk*mj" and use filter for source as Guru Granth Sahib ji
2. Scroll down to the last of the results, the last one isn't completely visible

theamanjs:
- [x] Fixes #1598 
larivaar assist fix on preview

- [x] Fixes #1599 
font inconsistency

- [x] Fixes #1600 
scrollability issue when filters applied

### DIFF
--- a/www/main/navigator/search/components/SearchContent.jsx
+++ b/www/main/navigator/search/components/SearchContent.jsx
@@ -148,7 +148,7 @@ const SearchContent = () => {
         <InputBox
           placeholder={getPlaceholder()}
           disabled={databaseProgress < 1}
-          className={`${currentLanguage === 'gr' && 'gurmukhi'} mousetrap`}
+          className={`${currentLanguage === 'gr' ? 'gurmukhi' : 'english'} mousetrap`}
           databaseProgress={databaseProgress}
           query={query}
           setQuery={setQuery}
--- a/www/main/navigator/search/components/SearchContent.jsx
+++ b/www/main/navigator/search/components/SearchContent.jsx
@@ -3,6 +3,7 @@ import { useStoreActions, useStoreState } from 'easy-peasy';
 import { remote, ipcRenderer } from 'electron';
 import banidb from '../../../common/constants/banidb';
 import { filters } from '../../utils';
+import { classNames } from '../../../common/utils';
 import {
   IconButton,
   InputBox,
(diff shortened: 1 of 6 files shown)


theamanjs:
@Gauravjeetsingh veerji can you review this one, so we can push the update further?

www/main/navigator/search/components/SearchContent.jsx
```diff
@@ -228,7 +229,7 @@ const SearchContent = () => {
           />
         </div>
       </div>
-      <div className="search-results">
+      <div className={classNames('search-results', isShowFiltersTag && 'filter-applied')}>
```
Gauravjeetsingh, line 232: @theamanjs ji, why do we need a function call here?

with function:
````
<div className={classNames('search-results', isShowFiltersTag && 'filter-applied')}>
````

Whereas, without function we can simply use the template string directly
````
<div className=`'search-results' ${isShowFiltersTag && 'filter-applied'}`>
````

Gauravjeetsingh (changes requested):

www/main/navigator/search/components/SearchContent.jsx
theamanjs: veerji classnames is a utility function that I created to avoid multiple conditions on an element and to avoid a major issue that comes while using the second approach you mentioned.

when we will use `isShowFilersTag && 'filters-applied'` and the condition is false it's gonna render a class named `false` and we don't need that. To avoid that false we will have to use an unwanted ternary operator.

Gauravjeetsingh: Got it. Good to go.

Gauravjeetsingh (approved):

Issue #1615, opened 2022-07-18
Update Bani Overlay when the Vaahiguroo slide or other non-Panktee slides are selected


saintsoldierx:
(Auto generated from slack)
Would it be possible to update OBS when the Vaahiguroo slide or other non-Panktee slides are selected?
https://khalis.slack.com/archives/CFS6E54PN/p1658115691739659

saintsoldierx:
<img width="1110" alt="image" src="https://user-images.githubusercontent.com/27791330/179612686-8b3b5749-ebcd-43a8-a5b2-adc966961890.png">

saintsoldierx:
I was watching the broadcast above and noticed this. The projector screens on the wall in the photo are STTM and I saw that the overlay didn't update when they would select the ਵਾਹਿਗੁਰੂ slide

PR #1629, opened 2022-07-27
fix: show misc slide text on bani overlay screen

Gauravjeetsingh:
Peek 2022-07-27 15-00

Fixes issue #1615

### DIFF
--- a/www/obs/index.html
+++ b/www/obs/index.html
@@ -8,7 +8,7 @@
     <script src="socket.io.js"></script>
 
     <script>
-      var previewData = {
+      const previewData = {
         English: 'If the Lord shows His Mercy, then I keep Him enshrined within my heart. ||1||',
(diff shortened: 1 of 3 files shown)


www/obs/index.html
```diff
@@ -88,6 +88,21 @@
           Line.Gurmukhi === '' && Line.English === '';
       };
 
+      var writeText = function(text, isGurmukhi, isAnnoucement) {
```
theamanjs, line 91: Veerji does it has to be `var`?

theamanjs (changes requested):

www/obs/index.html
Gauravjeetsingh: It does not, thank you! pushed the fix.

theamanjs (approved):

Issue #136, opened 2018-05-26
Selectable order of presentations lines


saintsoldierx:
Users should be able to select the order of lines presented on screen. Drag & drop.

prebhdevsingh:
What problem are we trying to solve with this? @saintsoldierx

saintsoldierx:
Let's say I want transliterations to show above translations, or the teeka to show below Gurbani for katha. Should be able to easily move the rows up and down and order as required. You could create your own STTM templates in STTM v2 in powerpoint to do the same.

saintsoldierx:
![image](https://user-images.githubusercontent.com/27791330/179150239-98ec1646-d51c-435d-a7d2-38f2fb77eeef.png)

Gauravjeetsingh:
Feedback from @saintsoldierx ji

"I’m a little confused as to the UI for being able to choose order of rows. I don’t think it should be on the “preview” in settings but more on the main window preview. Also, it would be better to have a dropdown on each row to select what the user wants to display there rather than having these up and down arrows…

Each of these should be like dropdowns. It also helps because someone might want to have translation in english, translation in spanish and transliteration without any teeka. The current way, i think, will not allow them to do that."

![image](https://user-images.githubusercontent.com/1990932/187149240-316e3651-c5a6-48b0-ab4c-d4022b752c92.png)

jagmit:
Gurfateh Ji, I want to emphasis that such a feature would be much appreciated. In our Gurdwara the canvas onto which sttm is projected is not fully visible from every angle of the darbaar. Being able to place the Gurbani on the bottom or center would ensure that everyone in the sangat would at least be able to read along the Gurbani 🙏🏽

Drag and drop in both settings and quick tools would of course be favourable, but I wouldn't mind a more rudimentary arrow button UI if this would make this feature land sooner.

cc @saintsoldierx

saintsoldierx:
> 

Thank you Jagmit, even with this feature, our plan is not to allow Gurbani to be moved below others. The other translations/transliterations would be orderable. Maybe we can include alternative layouts to have Gurbani placed lower on screen, but I still would prefer not to have any other language above.

PR #1631, opened 2022-08-05
Change order of lines in viewer

Gauravjeetsingh:
Fixes issue #136 

Pending:
- [ ] Fix the CSS of order buttons

### DIFF
--- a/www/main/common/store/GlobalState.js
+++ b/www/main/common/store/GlobalState.js
@@ -62,6 +62,31 @@ const GlobalState = createStore({
   navigator: createNavigatorSettingsState(navigatorSettings),
   viewerSettings: {
     quickTools: false,
+    slideOrder: ['translation', 'teeka', 'transliteration'],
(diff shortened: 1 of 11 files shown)


www/src/scss/_settings.scss
```diff
@@ -24,6 +24,63 @@
...
+        .left-control:hover {
+          background-color: rgba(0, 0, 0, 0.25);
+
+          img {
+            opacity: 1;
+          }
+        }
+
+        .right-control:hover {
+          background-color: rgba(0, 0, 0, 0.25);
```
theamanjs, line 64: Can these colors come from color variables file?

theamanjs (changes requested):

www/src/scss/_settings.scss
Gauravjeetsingh: Yes, good catch, this should be fixed now. thank you veerji!

theamanjs (approved):

Issue #1632, opened 2022-08-08
Font spacing issues while using ChromeCast


theamanjs:
From Slack
> Hi, just wanted to check whether the fix for this issue will be coming soon as I believe it had been going on for a while now. Whenever we Chromecast from the desktop app, the display of the Gurbani words are very close together and hard to read as compared to when using HDMI or the laptop display itself. Hoping to use this the next couple of days for a Kirtan event, so was hoping that the fix might be released soon!

![pxl_20220804_170210591 mp](https://user-images.githubusercontent.com/38748298/183350321-73441c72-f2c1-4a4a-81fa-c92911f55135.jpg)
While using chromecast

<br />
<img width="1299" alt="image (3)" src="https://user-images.githubusercontent.com/38748298/183350333-81d86001-5be0-44a3-b7eb-f7fb759c21e1.png">
While using HDMI

PR #1634, opened 2022-08-19
fix: spacing issue on chromecast window

Gauravjeetsingh:
- add a keybord shortcut to search for chromecast devices
We can now use Ctrl/Cmd + Alt + C to open the chromecast search window.

Fixes issue #1632


theamanjs (changes requested):
Veerji, since you have added a new shortcut for Chromecast. I think it should be added to LEGEND.md file

Gauravjeetsingh:
Thank you @theamanjs veerji, good catch, updated the legend.md file.

theamanjs (approved):

Issue #1636, opened 2022-08-22
Sundar Gutka Baanis are not resuming from last read verse in history


theamanjs:
From Slack:
> Many users complaining that when opening Asa Di Vaar (or probably other Sunder Gutka baanis too), it doesn’t go back to the line they were on if they select a Shabad in between.

PR #1637, opened 2022-08-22
Fix the active verse change when switching from baani to shabad

Gauravjeetsingh:
Peek 2022-08-22 19-08


Gauravjeetsingh:
fixes issue #1636

theamanjs (approved):

Issue #1659, opened 2022-09-07
Minimising Shortcut tray crashes app when in dark modes


MySmilingTurban:
Steps to reproduce:
1. Choose any dark mode
2. Toggle the shortcut tray, the app crashes

Try the same thing in light modes, works fine

saintsoldierx:
Verified issue exists on Mac

Gauravjeetsingh:
Closing this out, as its released to prod.

PR #1661, opened 2022-09-08
Fix the app crash on toggling shortcut bar in dark mode

Gauravjeetsingh:
Adding style to the toggle class caused the app to crash on closing shortcut bar. Changing the class fixed the issue.

Peek 2022-09-08 17-58

Fixes issue #1659


theamanjs (approved):
Smooth veerji! :raised_hands:

Issue #1658, opened 2022-09-07
Auto-scroll acts weird on long shabads/banis using space


MySmilingTurban:
Steps to reproduce:
1. search any long shabad say **p**anch **p**arvaan **p**anch **p**ardhan or anand sahib (bhog)
2. select first or last tuk as house tuk
3. click on the tuk at the other end
4. press space, notice how auto-scroll does the opposite of what's expected
it takes to the house panti when the highlight is on other pankti and vice versa

Gauravjeetsingh:
Closing this out, as its released to prod.

PR #1662, opened 2022-09-09
Fix the scroll issue when using the space shorcut

Gauravjeetsingh:
Peek 2022-09-09 13-02

Fixes issue #1658


theamanjs (approved):
Great work veerji :heart:

PR #1687, opened 2022-11-17
chore: disables the scss linting

Gauravjeetsingh:
There was some error with scss linting, that was causing the automatic build to break, testing that by removing linting for now.
We can re-enable this after the scss files are fixed as per linter in the later release


inderpreetsingh:
Merging this one to let the test run on dev branch.

PR #1689, opened 2022-11-24
Resolve the linting issues

Gauravjeetsingh:
With upgrading the package, the eslint added new rules, which our code wasn't following. In this PR it resolves all the linting errors, or escape the ones which we intentionally didn't follow.


.eslintrc
```diff
@@ -11,7 +11,8 @@
     "prettier/prettier": ["error"],
     "import/prefer-default-export": "off",
     "no-plusplus": "off",
-    "class-methods-use-this": "off"
+    "class-methods-use-this": "off",
+    "no-restricted-exports": "off"
```
Gauravjeetsingh, line 15: This triggered the modules we exported with default as
export { default } from 'abc/xyz.js'

We were using this pattern at multiple times, so escaped this for now. 
Let me know if I am understanding it wrong, here's the linting rule details: https://eslint.org/docs/latest/rules/no-restricted-exports

package.json
```diff
@@ -80,6 +79,7 @@
     "copy-to-clipboard": "^3.3.1",
     "dotenv": "^16.0.1",
     "easy-peasy": "^5.0.4",
+		"electron": "^19.0.8",
```
Gauravjeetsingh, line 82: ESLint said to list electron as dependency instead of dev-dependency, so I manually changed its position in package.json.

www/main/common/store/user-settings/get-saved-user-settings.js
```diff
@@ -3,14 +3,17 @@ import convertObjToCamelCase from '../../utils/convert-object-to-camel-case';
...
 const { settings } = require('../../../../configs/user-settings.json');
 
 let userDataPath;
 
 if (electron.app) {
   userDataPath = electron.app.getPath('userData');
 } else {
-  const app = require('@electron/remote').app;
+  /* eslint-disable global-require */
+  const { app } = require('@electron/remote');
```
Gauravjeetsingh, line 15: This exception is used at multiple places through out the code. Linting rules want require to be the first thing in code, but we are using require in conditionals or with some code logic like above.

www/main/navigator/search/utils/retrieve-from-db.js
```diff
@@ -1,15 +1,15 @@
...
 const { i18n } = remote.require('./app');
 
 export const retrieveFilterOption = async (optionsObj, type) => {
   const idArray = Object.keys(optionsObj).filter(
-    option => option.toLowerCase() !== 'all' && option.toLowerCase() !== 'others',
+    (option) => option.toLowerCase() !== 'all' && option.toLowerCase() !== 'others',
   );
   const retrievedObj = await getFilterOption(type, idArray);
-  const valueObj = Object.assign({}, retrievedObj);
+  const valueObj = { ...retrievedObj };
```
Gauravjeetsingh, line 12: @theamanjs veerji, can you confirm if this is correct, the linter did it on its own. I checked the functionality and the app is working fine.

www/main/viewer.js
```diff
@@ -1,339 +0,0 @@
-/* eslint
```
Gauravjeetsingh, line 1: This code was not being used, and was from the old code, so deleted entire file.

www/main/viewer/utils/chromecast.js
```diff
@@ -1,182 +1,192 @@
-const remote = require("@electron/remote");
+/* global chrome */
+import { ipcRenderer } from 'electron';
+
+const remote = require('@electron/remote');
+
+/* eslint-disable global-require */
+export const tingle = require('../../../assets/js/vendor/tingle');
```
Gauravjeetsingh, line 7: Linter didnt allow exporting a module imported with require. But this is being used at multiple places so I added an exception.

www/main/viewer/utils/chromecast.js
```diff
@@ -1,182 +1,192 @@
-const remote = require("@electron/remote");
+/* global chrome */
```
theamanjs, line 1: Veerji, I'm not sure if I'm familiar with this one. Can you help me know this one?

www/main/viewer/utils/chromecast.js
theamanjs: Okay veerji. But you can just disable this for just next line.

www/main/viewer/utils/chromecast.js
```diff
@@ -1,182 +1,192 @@
...
 let isCastInitialized = false;
 
+/**
+ * append message to debug message window
+ * @param {string} message A message string
+ */
+export const appendMessage = (message) => {
+  /* eslint-disable no-console */
+  console.log(message);
+  /* eslint-enable */
```
theamanjs, line 25: Veerji, this one too can be just disabled for just next line.

www/main/viewer/utils/chromecast.js
```diff
@@ -1,182 +1,192 @@
...
+    closeMethods: ['overlay', 'button', 'escape'],
+  });
+
+  modal.setContent(`<h2>${errorMessage}</h2>`);
+  // add ok button
+  modal.addFooterBtn('OK', 'tingle-btn tingle-btn--pull-right tingle-btn--default', () => {
+    modal.close();
+  });
+
+  // open modal
```
theamanjs, line 41: Veerji, I think these obvious comments can be removed. However, I think this is a little subjective thing. What do you say?

theamanjs (changes requested):
Veerji, there are some issues that are identical to each other but I have comment everywhere just to make sure we don't miss any.

www/main/viewer/utils/chromecast.js
Gauravjeetsingh: @theamanjs chrome object is imported after a script is initialized. If we don't specify this linter will give an undefined error, because the object is intialized on run time.
https://eslint.org/docs/latest/user-guide/configuring/language-options#specifying-globals

www/main/viewer/utils/chromecast.js
Gauravjeetsingh: I think this makes it easier to find code, we can search for open modal to see this entire code. Similarly searching for ok button will show us that button code.

Gauravjeetsingh:
@theamanjs ji done with the changes, have a look!

www/main/viewer/utils/chromecast.js
theamanjs: Thankyou veerji. This is helpful :raised_hands:

www/main/viewer/utils/chromecast.js
theamanjs: Okay veerji. I was thinking these comments are too obvious for this piece of code. However, we can keep this as it is to find code easily.

theamanjs:
LGTM @Gauravjeetsingh veerji

theamanjs (approved):

PR #1705, opened 2023-01-04
Fix the build error and custom background deletion bug

Gauravjeetsingh:
Currently, it was resetting the background even if we delete an inactive custom background, fixed that.

Also, resolved linting errors and remove unused code.

### DIFF
--- a/www/main/settings/components/ThemeContainer.jsx
+++ b/www/main/settings/components/ThemeContainer.jsx
@@ -81,8 +81,12 @@ const ThemeContainer = () => {
                 onRemove={() => {
                   removeCustomBackgroundFile(tile['background-image-path'].replace(/\\(\s)/g, ' '));
                   upsertCustomBackgrounds(setCustomThemes);
-                  const currentThemeInstance = themes.filter((theme) => theme.key === currentTheme);
-                  setDefaultBg(currentThemeInstance, setThemeBg, themeBg);
+                  if (tile['background-image'] === themeBg.url.href) {
+                    const currentThemeInstance = themes.filter(
+                      (theme) => theme.key === currentTheme,
+                    );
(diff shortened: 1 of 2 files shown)


www/main/settings/components/ThemeContainer.jsx
```diff
@@ -81,8 +81,12 @@ const ThemeContainer = () => {
                 onRemove={() => {
                   removeCustomBackgroundFile(tile['background-image-path'].replace(/\\(\s)/g, ' '));
                   upsertCustomBackgrounds(setCustomThemes);
-                  const currentThemeInstance = themes.filter((theme) => theme.key === currentTheme);
-                  setDefaultBg(currentThemeInstance, setThemeBg, themeBg);
+                  if (tile['background-image'] === themeBg.url.href) {
+                    const currentThemeInstance = themes.filter(
+                      (theme) => theme.key === currentTheme,
+                    );
+                    setDefaultBg(currentThemeInstance, setThemeBg, themeBg);
```
theamanjs, line 88: The scope of `currentThemeInstance` variable is inside the if statement only and it is being used outside of that if statement which is a violation. Veerji this should be fixed IMO

theamanjs (changes requested):

www/main/settings/components/ThemeContainer.jsx
Gauravjeetsingh: It is inside if statement, if statement, ends at line 89.

theamanjs: I messed up veerji @Gauravjeetsingh really sorry for this 🙏

theamanjs (approved):

theamanjs:
@inderpreetsingh veerji this PR is ready to be merged after your review.

inderpreetsingh (approved):

Issue #1727, opened 2023-02-11
Break issue in Asa Di Vaar, jumps position


saintsoldierx:
See video for demo. 

https://user-images.githubusercontent.com/27791330/218276678-259593d9-da2c-495b-917d-efbc8b25aff9.mp4

PR #1750, opened 2023-04-27
Fix asa di vaar unwanted scroll issue

theamanjs:
This PR fixes issue #1727 
- Also, fixed the stutter on changing verse

https://user-images.githubusercontent.com/38748298/235840008-c3e58cce-69d9-414f-aa3d-a3653175babf.mov

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "sttm-desktop",
-  "version": "9.1.0",
+  "version": "9.1.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "sttm-desktop",
-      "version": "9.1.0",
+      "version": "9.1.1",
       "hasInstallScript": true,
       "license": "OSL-3.0",
       "dependencies": {
--- a/www/main/navigator/shabad/ShabadContent.jsx
+++ b/www/main/navigator/shabad/ShabadContent.jsx
@@ -362,12 +362,15 @@ const ShabadContent = () => {
 
   const scrollToView = () => {
     setTimeout(() => {
-      const currentIndex = activeShabad.findIndex((obj) => obj.ID === activeVerseId);
-      virtuosoRef.current.scrollToIndex({
-        index: currentIndex,
-        behavior: 'smooth',
-        align: 'center',
-      });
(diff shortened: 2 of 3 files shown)


www/main/navigator/shabad/ShabadContent.jsx
```diff
@@ -148,6 +170,7 @@ const ShabadContent = () => {
     }
     setActiveVerse({ [verseIndex]: newTraversedVerse });
     if (activeVerseId !== newTraversedVerse) {
+      console.log('newTraversedVerse', newTraversedVerse);
```
Gauravjeetsingh, line 173: @theamanjs veerji, we can remove this console log here, I believe.

www/main/navigator/shabad/ShabadContent.jsx
```diff
@@ -217,32 +239,14 @@ const ShabadContent = () => {
...
-    virtuosoRef.current.scrollToIndex({
-      index: verseIndex,
-      behavior: 'smooth',
-      align: 'center',
-    });
-  };
-
   const skipMangla = (shabadVerses, index) => {
-    const gurmukhi = shabadVerses[index]?.verse;
+    const gurmukhi = shabadVerses[index].verse;
```
Gauravjeetsingh, line 243: @theamanjs Removing ? might cause error, because sometimes there is no verse property, So we need a conditional to set it to gurmukhi only when there is verse property.

package-lock.json
Gauravjeetsingh, line 1: @theamanjs how come package-lock file is changed with this issue, did we add any new package or update any previous package?

Gauravjeetsingh (changes requested):

package-lock.json
theamanjs: No veer ji but the package-lock did now have an updated application. This might be due to package-lock file not being updated in the last release. This time I ran npm install and it updated the package-lock file.

www/main/navigator/shabad/ShabadContent.jsx
theamanjs: Fixed this one veerji, thanks for pointing out.

www/main/navigator/shabad/ShabadContent.jsx
theamanjs: fixed

Gauravjeetsingh (approved):

Issue #1720, opened 2023-02-08
Need more search results


saintsoldierx:
The limit of 20 search results for a given query is not enough. There needs to be a "more" button that pops up a new window with all search results (maybe in infinity scroll) that also allows to filter. When clicking a shabad, it should then close and display it.

PR #1755, opened 2023-05-16
Add infinite scroll to search result

theamanjs:
## Add enhancement #1720 
On scrolling to the end, it will keep adding 20 results to the bottom

https://github.com/KhalisFoundation/sttm-desktop/assets/38748298/eed231ae-b567-4bb8-960c-f9d81148cf8b

### DIFF
--- a/www/main/banidb/realm-search.js
+++ b/www/main/banidb/realm-search.js
@@ -49,7 +49,7 @@ const init = () => {
  * search('jggsspp', 0, 'all');
  * // => [{ Gurmukhi: 'jo gurisK guru syvdy sy puMn prwxI ]', ID: 31057 },...]
  */
-const query = (searchQuery, searchType, searchSource) =>
+const query = (searchQuery, searchType, searchSource, howManyRowsToShow = 20) =>
   new Promise((resolve, reject) => {
     if (!initialized) {
       init();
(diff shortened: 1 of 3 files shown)


www/main/banidb/realm-search.js
```diff
@@ -49,7 +49,7 @@ const init = () => {
  * search('jggsspp', 0, 'all');
  * // => [{ Gurmukhi: 'jo gurisK guru syvdy sy puMn prwxI ]', ID: 31057 },...]
  */
-const query = (searchQuery, searchType, searchSource) =>
+const query = (searchQuery, searchType, searchSource, howManyRowsToShow = 20) =>
```
Gauravjeetsingh, line 52: Not really a bug, but was thinking if we can we rename this state name to something like `resultRows` ? It would be more clearer imo.

www/main/banidb/realm-search.js
```diff
@@ -62,7 +62,7 @@ const query = (searchQuery, searchType, searchSource) =>
     // default source for ang search to GURU_GRANTH_SAHIB
     let angSearchSourceId = CONSTS.SOURCE_TYPES.GURU_GRANTH_SAHIB;
     const order = [];
-    let howManyRows = 20;
+    let howManyRows = howManyRowsToShow;
```
Gauravjeetsingh, line 65: @theamanjs veerji 
where are we using this variable, can't we use the state directly?

Gauravjeetsingh (changes requested):
Added few comments in code, have a look @theamanjs veerji

www/main/banidb/realm-search.js
theamanjs: Veerji, I'm using this because we cannot mutate the argument passed to a function. Eslint would throw error on param reassign.

The variable is being reassigned a value here https://github.com/KhalisFoundation/sttm-desktop/blob/2f91ec203077cfe7c5d010d3fb4ef75741607249/www/main/banidb/realm-search.js#L131

www/main/banidb/realm-search.js
theamanjs: Done veerji

Gauravjeetsingh (approved):

PR #1760, opened 2023-05-25
1724/dhan guru overhaul

### LINKED ISSUE #1724: Improvements to Dhan Guru
1. Rename the tab "insert" to "Gurus"
2. Include English transliteration like "Dhan Guru Gobind Singh Sahib Jee"
3. Center the text in middle (horizontally) not at the top.
4. Re-do the tab pane as per screenshot

image

theamanjs:
- Fix #1724


www/main/navigator/misc/components/DhanGuruPane.jsx
```diff
@@ -0,0 +1,201 @@
...
+              insertDhanGuru(index);
+            }}
+            onMouseEnter={() => {
+              refsByGuruIndex[guru].current.classList.add('dhan-guru-button-prefix-hover');
+              refsByGuruIndex[guru].current.innerHTML = '';
+            }}
+            onMouseLeave={() => {
+              refsByGuruIndex[guru].current.classList.remove('dhan-guru-button-prefix-hover');
+              refsByGuruIndex[guru].current.innerHTML = getGuruIndex(index);
+            }}
```
Gauravjeetsingh, line 186: @theamanjs veerji, I don't think we are using dhan-guru-button-prefix-hover class when styling. It styles directly using hover selector. Is there any other significance of this class?

Also, instead of adding an event here, how about adding an element in span, and setting it to display none on hover?

```
<span className="dhan-guru-button-prefix" ref={refsByGuruIndex[guru]}>
    <span className="prefix-content">  {getGuruIndex(index)} </span>
  </span>
```

In misc.scss, hide prefix-content, when hovered on parent div.

We won't need those events and the references here in that case.

Gauravjeetsingh (changes requested):

Gauravjeetsingh:
@theamanjs Also, with this change, there's no way to add the dhan guru slides from single display mode.

Gauravjeetsingh:
Also, I missed another side effect of css change, can you fix that as we did the dhan guru pane height.

In single display mode, we get two scrollbars now.
image

www/main/navigator/misc/components/DhanGuruPane.jsx
theamanjs: I have removed that mouse event code and applied color transparent to the number on hover. You can review that.

theamanjs:
@Gauravjeetsingh veerji you can check all the issues that you mentioned are now fixed.

Gauravjeetsingh:
@theamanjs veerji
There are two scrollbars in shabad pane in single display mode. Can you fix it the same way.
Screenshot from 2023-06-02 11-08-48

Also, the current height of search results seemed a little less on my laptop screen (1366 x 768), what if we set a min-height to this complete pane, which would give more space to the search results, automatically.

Screenshot from 2023-06-02 11-09-00

theamanjs:
@Gauravjeetsingh veerji I have opened a new issue and this can be done in separate PR https://github.com/KhalisFoundation/sttm-desktop/issues/1765 Let's see if anything directly related to dhan guru is giving issues and can be fixed in this PR

Gauravjeetsingh:
@theamanjs there was a slight CSS change required, I pushed that, all good from my side now. Should be good to merge.

Gauravjeetsingh (approved):

PR #1766, opened 2023-06-07
fix: search result size when filter applied

theamanjs:
Fix this issue of size when filter is applied 
image


theamanjs:
@inderpreetsingh veerji can you quickly review this as well? before merging 9.1.2?

inderpreetsingh (approved):

Issue #1776, opened 2023-06-21
9.1.3 (MacOS) - Dhan Guru slides and Akhand Paath View combo error


MySmilingTurban:
Steps to reproduce:
- Go to any shabad and navigate using Akhand Paath View
- Show a guru slide, notice it shows repeated name of Guru sahib
- Also, when you go back to shabad, it gets stuck (navigation doesn't work)

theamanjs:
Moving this issue to 9.1.3 Since we gotta release 9.1.2 and it's the priority

Gauravjeetsingh:
This has been pushed to production, closing this out.

PR #1788, opened 2023-06-28
9.1.2 fix: repeated slides when akhand paath mode is on

theamanjs:
fixes issue #1776 
- Yet to fix the css issue of verse going on top of viewer


theamanjs:
Discarding this PR since it's creating other issues related to Akhand Paath view. Should be done in next 9.1.3 PR. 9.1.2 is priority at this moment. #1776

PR #1816, opened 2023-08-08
Show english transliteration in search result

Gauravjeetsingh:
image

### DIFF
--- a/www/main/common/sttm-ui/search-results/SearchResults.jsx
+++ b/www/main/common/sttm-ui/search-results/SearchResults.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import anvaad from 'anvaad-js';
 
 const remote = require('@electron/remote');
+
 const { i18n } = remote.require('./app');
 
 const SearchResults = ({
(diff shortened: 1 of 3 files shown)


theamanjs:
@Gauravjeetsingh I believe if we capitalize the verse it would look better than now.

www/main/common/sttm-ui/search-results/SearchResults.jsx
```diff
@@ -107,6 +109,7 @@ const SearchResults = ({
             )} ${ang} `}</span>
           )}
           <span className="gurmukhi">{highlightKeywords(verse, searchQuery)}</span>
+          {currentLanguage === 'en' && <div className="eng-verse">{anvaad.translit(verse)}</div>}
```
theamanjs, line 112: veerji, recently I learned this. We should keep our static strings in a standard file. Just like 'en' all the strings from all the files should be moved to a javascript file. You can create this issue and work on this later.

www/src/scss/navigator/common/_verse-block.scss
```diff
@@ -88,12 +88,17 @@
...
-
     div {
       display: inline-flex;
       a {
         padding: 5px;
         font-size: 1rem;
+
+        .eng-verse {
+          color: $rust;
+          display:block;
```
theamanjs, line 99: indentation issue

theamanjs (changes requested):

www/src/scss/navigator/common/_verse-block.scss
Gauravjeetsingh: Looks fine to me veerji, it has two spaces everywhere.

Issue #1488, opened 2022-03-21
Larivaar assist and vishraams are missing in bani overlay mode


theamanjs:
No Larivaar Assist or Vishraams showing in Bani Overlay Mode.

PR #1842, opened 2023-11-13
Add larivaar settings in Bani Overlay

Gauravjeetsingh:
Add toggle switch for Larivaar and Larivaar Assist setting
larivaar

Fixes issue #1488


inderpreetsingh (changes requested):
@Gauravjeetsingh Can you reduce the size of toggle switch  below larivaar and larivaar assist. Also, maybe reduce the space between both options.

Gauravjeetsingh:
Done

Issue #1566, opened 2022-05-06
Make highlighted text heavier


saintsoldierx:
Cannot really tell easily between the highlighted text when using search. I would suggest to make the highlighted text more heavy/bold. It's not too bad in light mode but dark mode is much more tougher to differentiate.

<img width="708" alt="image" src="https://user-images.githubusercontent.com/27791330/167089068-be746787-dc02-44bc-a9ee-7f1457ba1c93.png">

<img width="687" alt="image" src="https://user-images.githubusercontent.com/27791330/167089249-74f7b29c-8ebe-489d-ae8c-99f1df8daa36.png">

theamanjs:
This got reported on slack again on dark mode. Will be working on this in upcoming releases. Message link: https://khalis.slack.com/archives/CFS6E54PN/p1663306180980609

Issue #1317, opened 2021-03-02
Animations


ManjotS:
<div> Freshdesk Ticket ID: <a href="https://khalis.freshdesk.com/a/tickets/3928">3928</a></div>
                       <div> Freshdesk Ticket Agent: Aman Singh</div>
                       <div> Freshdesk Ticket Agent Email: saintsoldierx@gmail.com</div>
                       <div> Ticket Priority: Low</div>
                       <div> Freshdesk Ticket Description: </div><div>From: Arshdeep Singh &lt;sarshdeep825@gmail.com&gt;<br>Subject: Suggestions about SikhiToTheMax<br>IP: 157.36.187.165 <br><br>Message Body:<br>Waheguru Ji Ka Khalsa, Waheguru Ji Ki Fateh Ji!<br>Your app (SikhiToTheMax) is great! I started using this in 2020. I would like to give you a suggestion about the app (PC) for the presentations/overlay, that when we change a specific line in a shabad, it changes with no animation. It would be great if you add fade animation while changing line in shabad. I shall be thankful to you!<br>Arshdeep Singh<br>Waheguru Ji Ka Khalsa, Waheguru Ji Ki Fateh!<br><br>--<br>This e-mail was sent using the contact form on Khalis Foundation (<a href="http://khalisfoundation.org" rel="noreferrer">http://khalisfoundation.org</a>)<br><br>
</div>

ManjotS:
<div>Agent Aman Singh linked Freshdesk ticket <a href="https://khalis.freshdesk.com/a/tickets/3928">3928</a> for this issue.</div>

Gauravjeetsingh:
Duplicate of #258 closing this out

Issue #1155, opened 2020-06-25
Next Prev Bugs


inderpreetsingh:
* If we use next and previous icon on shabad pane, in sangat sync it doesn't change the shabad.
* Next / Prev should be hidden when opened bani or ceremonies
* I am not sure if we wanted that as a behaviour, but pointing out (just in case). When we change the Shabad from next previous icons in shabad pane, that history is not saved in history pane.

PR #1845, opened 2023-11-17
Minor Enhancements

Gauravjeetsingh:
Fixes issue #1155 
- Hide the next/prev icon in Sundar gutka bani or ceremony

Fixes issue #1566 
- Make the highlighted text stand out
image
image

Fixes issue #1317 
- Add fade transition in viewer when changing verse (screen recorder didn't catch the smooth effect, its more smoother in actual)
animation


saintsoldierx:
In regards to the highlighting - I would leave the background transparent unless you can add some left and right padding to each word about 5px. Another option would be is to highlight those words by changing color of the font to some sort of yellow?

saintsoldierx:
Can the fade transition be crossfade rather than blanking to white and fading in every time?

Gauravjeetsingh:
@saintsoldierx veerji, removed the background color from highlighted words. 

Crossfade won't be possible with current setup, because the viewer only has the active verse, which is changed on clicking. We would have to redo that logic to have both previous and active verse in component for crossfade to work.

Gauravjeetsingh:
Merging this for alpha release, will fix the fade animation as discussed on slack in a separate PR

Issue #1121, opened 2020-06-11
Reset Font Sizes 


maneetpaul:
An option to reset all font sizes to default. This will only impact font sizes and no other setting.

It should be in the settings pane in the first section as the last option below the font sliders

PR #1906, opened 2024-03-13
Implement Reset Font Sizes Option

yadvirkaur:
#1121


www/main/common/store/user-settings/create-user-settings-state.js
```diff
@@ -16,6 +16,45 @@ const createUserSettingsState = (settingsSchema, savedSettings, userConfigPath)
       userSettingsState[stateVarName] = savedSettings[settingKey];
     }
 
+    // Defines an action to reset all font size settings to their initial values.
+    userSettingsState.resetFontSizes = action((state) => {
```
Gauravjeetsingh, line 20: We shouldn't have to rewrite all the logic here. This is automatically generated if we add a new setting in user-settings.json file.

So the flow would be
- Add a setting in user-settings.json in category `font-sizes` In file user-settings.json
```json
 "font-sizes": {
      "title": "FONT_SIZES",
      "type": "range",
      "max": 20,
      "min": 1,
      "step": 1,
      "settings": [
        "gurbani-font-size",
        "content1-font-size",
        "content2-font-size",
        "content3-font-size",
        "announcements-font-size",
        <add-new -setting-here>
      ]
    },
```

- In same file, inside settings object, add a new setting for reset button, you can add a new type for this, let's call it reset-button, and another property named resetSettings, that will give an array of settings it will reset.
```
"settings": {
    "<new-setting-here>": {
      "title": "SOME NAME HERE",
      "type": "reset-button"
      "resetSettings": ['gurbani-font-size', 'content1-font-size', 'content2-font-size', 'content3-font-size', 'announcement-font-size']
    },
```

- In Setting.jsx, add a new type inside switch statement for this, where you add a button and reset these settings using onclick attribute.

To understand more about how its working, read through this [documentation](https://github.com/KhalisFoundation/sttm-desktop/wiki/Settings-Doc-Part-1:-The-structure-of-user-settings.json-(feat-basics-of-global-state-management)#structure-of-user-settingsjson) and we can discuss if you have any doubts.

Gauravjeetsingh (changes requested):

www/main/settings/components/Setting.jsx
```diff
@@ -131,6 +144,17 @@ const Setting = ({ settingObj, stateVar, stateFunction }) => {
...
       );
       break;
+    case 'reset-button':
+      settingDOM = (
+        <button
+          onClick={handleResetFontSizes}
+          className="icon-reset"
+          title={settingObj.tooltip}
+          style={{ backgroundImage: `url('assets/img/icons/${settingObj.icon}')` }}
+        >
```
Gauravjeetsingh, line 154: Add the text "Reset to Default"alongwith the icon in button. Instead of icon as background image, then you can add it as children element of button tag. 

Also, set cursor of button to pointer in css file.

Gauravjeetsingh (changes requested):
All good @yadvirkaur ji, a minor style change, and we would be good to merge this PR.

www/main/settings/components/Setting.jsx
```diff
@@ -131,6 +144,17 @@ const Setting = ({ settingObj, stateVar, stateFunction }) => {
         </>
       );
       break;
+    case 'reset-button':
+      settingDOM = (
+        <div
```
Gauravjeetsingh, line 149: @yadvirkaur Can we use a button tag here, I think that fits more nicely with the setting type.

Gauravjeetsingh (changes requested):

www/locales/en.json
```diff
@@ -301,6 +301,8 @@
     "PUNJABI_TEEKA": "Punjabi Teeka",
     "TRANSLATION": "Translation",
     "TRANSLITERATION": "Transliteration",
+    "RESET FONT SIZES": "Reset Font Sizes",
```
Gauravjeetsingh, line 304: @yadvirkaur Can you please use underscores instead of spaces in the key. We have that standard followed throughout this file.

It would become: RESET_FONT_SIZE, and so on.

Gauravjeetsingh (changes requested):
A minor change needed, then we can merge it. Thank you for the fix.

www/locales/en.json
yadvirkaur: Done. Thank you for noticing, it just skipped.

Issue #1193, opened 2020-08-25
Bani Overlay copy should give feedback


saintsoldierx:
When clicking “Copy” on URL for Bani Overlay, it should have a message or some feedback to tell you its been copied. Clicking on the URL should also copy it to your clipboard.

PR #1908, opened 2024-03-18
Implement feedback on copy for Bani Overlay URL

yadvirkaur:
#1193 

Bani Overlay URL copy functionality:
- When clicking on the copy icon for the Bani Overlay URL, it copies the URL to your clipboard and displays feedback. 
- Clicking on the URL selects URL text and copies it to your clipboard.

### DIFF
--- a/www/main/overlay/components/CopyToClipboard.jsx
+++ b/www/main/overlay/components/CopyToClipboard.jsx
@@ -6,22 +6,42 @@ const copy = require('copy-to-clipboard');
 
 const { i18n } = remote.require('./app');
 
-export const CopyToClipboard = ({ url }) => (
-  <div className="overlay-url">
-    <p className="overlay-window-text live-url-header">{i18n.t(`BANI_OVERLAY.LIVE_URL`)}</p>
-    <div className="url-container">
-      <input disabled={true} type="text" className="url-text" value={url} />
-      <span
-        className="export-btn"
-        onClick={() => {
-          copy(url);
-        }}
-      >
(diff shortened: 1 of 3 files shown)


www/main/overlay/components/CopyToClipboard.jsx
```diff
@@ -6,22 +6,42 @@ const copy = require('copy-to-clipboard');
...
-      >
-        <i className="fa fa-files-o cp-icon"></i>
-      </span>
+export const CopyToClipboard = ({ url }) => {
+  const [copied, setCopied] = React.useState(false);
+
+  const handleCopy = () => {
+    copy(url);
+    setCopied(!copied);
+    setTimeout(() => setCopied(false), 2000); // Reset the copied state after 2 seconds.
```
Gauravjeetsingh, line 15: Instead of doing `setTimeout` what if we use the mouseleave event?
https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseleave_event

So the flow would be, 
- by default the value of tooltip is **Copy URL**.
- After click it changed to **Copied!**
- On mouseleave event from that copy icon, change tooltip value back to Copy URL.

Currently, it shows Copied, and if I keep my mouse cursor there, it changes back to Copy URL.

overlay-copy

Gauravjeetsingh (changes requested):
Just one minor change, everything else looks good.

Gauravjeetsingh (approved):
All good, merging this PR. Thank you @yadvirkaur

inderpreetsingh (approved):

Issue #1909, opened 2024-03-23
Logo display in Bani Overlay window doesn’t seem to work


saintsoldierx:
(Auto generated from slack)
Logo display in Bani Overlay window doesn’t seem to work
https://khalis.slack.com/archives/CFS6E54PN/p1711171531082549

Gauravjeetsingh:
Fixed and released to production.

PR #1914, opened 2024-03-27
Fix: Ensure consistent logo display across Bani Overlay layouts

yadvirkaur:
#1909


Gauravjeetsingh (changes requested):
Looks good, minor change, Can we use the logo we have in split display, attaching the screenshot below.
image

For layouts, in which the logo is outside the content box, use the css filter property to add a subtle drop shadow to it so that it stays visible.

Gauravjeetsingh:
@yadvirkaur The logo is still old one in other layouts
image


Can we use the white transparent logo in every layout, and in top and bottom layouts, adding a shadow to the image, so that it stays visible.

I am thinking of making it something like this
image

Gauravjeetsingh (changes requested):
Need to update logo in all layouts

yadvirkaur:
Yes, will update this. Previously I kept the transparent logo only where it was inside the box.

Gauravjeetsingh (approved):

Issue #1802, opened 2023-07-07
Code Clean up and minor refactor


theamanjs:
# Code cleanup tasks

- [ ] Should add proptypes to the files where it’s missing
- [ ] File: SearchContent.jsx line:62 we are using static string ‘all’ instead that should some from banidb.js file. Should replace the static strings at all the places.
- [ ] File: MiscHeader.jsx line: 51 we are using static string this should some from i18n. Issues like that should be fixed, not limited to this file only.
- [ ] File: MiscHeader.jsc line line: 36 we are passign static string to setTab file. Instead there should be a common js file in constants folder which should have all these tabs name and replace that variable in all the files that are using this static string. Not limited to the mentioned line number and file.
- [ ] File: shortcut-functions.js is empty. Should be removed.
- [ ] File: shortcut-map.json is no longer in use. Either we should find a way to use that again by separating the keyboard shortcut code from Launchpad.jsx file or remove it
- [ ] File: shortcut_tray.json is not in use. Find a way to get that in use or remove it
- [ ] File: viewer.json is empty. Remove or get in use
- [ ] File: .prettierrc maybe not in use. Get that in use again
- [ ] We can’t see new changes, without shutting down the server and starting again, it shouldn’t be this way. Some problem with how source maps are integrated; hot reload

PR #1923, opened 2024-04-15
Minor Refactor: Move tab names into a common js file

yadvirkaur:
#1802

This solution is for the task: "File: MiscHeader.jsx line: 36 we are passing a static string to setTab. Instead, there should be a common js file in constants folder which should have all these tab names and replace that variable in all the files that are using this static string. Not limited to the mentioned line number and file."

Solution: Move all tab names to a new constants file in the constants folder and replace all static string references to tab names across the application with these constants.

### DIFF
--- a/www/main/common/constants/misc-tabs.js
+++ b/www/main/common/constants/misc-tabs.js
@@ -1,0 +1,9 @@
+const TAB_NAMES = {
+  HISTORY: 'History',
+  FAVORITES: 'Favorites',
+  OTHERS: 'Others',
+};
+
+module.exports = {
+  TAB_NAMES,
+};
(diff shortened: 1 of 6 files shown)


yadvirkaur:
I did this for only the History, Favourites and Others tabs, kindly review the solution, if I am in the right direction this needs to be done for other tabs too such as : Shabads, Search, Announcements and Dhan-Guru. SImilar thing can be done for overlay screens too. Guide me here.

Gauravjeetsingh:
@yadvirkaur as we discussed this in call, can we instead use the en.json file for constants, instead of a separate file here.

yadvirkaur:
I began with en.json but realised these strings are not user-facing, but are being used for internal logic in code. Using en.json might introduce risks of translation affecting the application's logic. How should we proceed?

Gauravjeetsingh:
Let me think a little on this one, I will get back to you on this in a day or two.

Gauravjeetsingh:
Closing this out, we can have another look here in next possible refactor sprint.

Issue #1761, opened 2023-05-29
Add setting to turn off intelligent spacebar


saintsoldierx:
When off, it should always just go to home pankti

Gauravjeetsingh:
@saintsoldierx veerji, we can definetly do that, but I am a little hesistant in adding a setting for this. If the spacebar is not being used intelligently, maybe have two separate shortcuts? One for intelligent scrolling and keep space bar for home pankti?

It could be **shift + space**, its intelligent scrolling
or any other single key event, like **i**

saintsoldierx:
No, i disagree - reason being is that spacebar is the most easiest key on the keyboard to use and our users are not only used to it, but it seems to be more intuitive than any other key. So i would say to add a setting for this and give users the option to switch.

singh-sp:
Waheguru ji,

I have the change ready that allows users to disable the (not so) "intelligent spacebar" feature. Note: I have left it enabled by default.

I am sure it has caused many others such a hassle since it was released. 😅- encountered numerous bugs (I was too lazy to report them).

Preview:

https://github.com/KhalisFoundation/sttm-desktop/assets/6597802/fac72d13-9885-4a27-9616-bd39e39330e6

saintsoldierx:
@singh-sp thank you, can you create a PR? I am wondering though why you don't like the intelligent spacebar feature? It's designed for Keertan where you can use it auto switch from asthaees to antras and back again easily.

singh-sp:
> @singh-sp thank you, can you create a PR? I am wondering though why you don't like the intelligent spacebar feature? It's designed for Keertan where you can use it auto switch from asthaees to antras and back again easily. 

Sure! I will create a PR tomorrow.

The reason we (myself and others) don't like it is because it forces you to follow a specific order. In reality, a Raagi can sing a shabad in *any order and may choose to revisit some panktis. When that happens, the spacebar becomes completely useless and out of sync, which is incredibly frustrating!

saintsoldierx:
@singh-sp sure i get that. But surely it works in many situations too? Would it be beneficial then to have a shortcut as well to enable/disable it rather than going to settings each time?

singh-sp:
Perhaps it works for some folks, and that’s great!

[...]

PR #1960, opened 2024-06-02
feat: add setting to turn off intelligent spacebar

singh-sp:
#1761 Intelligent spacebar can now be turned on/off in settings.

> Note: It will remain enabled for existing users.

Preview:

https://github.com/KhalisFoundation/sttm-desktop/assets/6597802/fac72d13-9885-4a27-9616-bd39e39330e6

### DIFF
--- a/www/configs/user-settings.json
+++ b/www/configs/user-settings.json
@@ -73,7 +73,7 @@
     "other-options": {
       "title": "OTHER_OPTIONS",
       "type": "switch",
-      "settings": ["limit-change-log", "statistics"]
+      "settings": ["limit-change-log", "statistics", "intelligent-spacebar"]
     }
   },
   "settings": {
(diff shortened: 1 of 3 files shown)


www/configs/user-settings.json
```diff
@@ -73,7 +73,7 @@
     "other-options": {
       "title": "OTHER_OPTIONS",
       "type": "switch",
-      "settings": ["limit-change-log", "statistics"]
+      "settings": ["limit-change-log", "statistics", "intelligent-spacebar"]
```
Gauravjeetsingh, line 76: Small change, Instead of other options, can you please move this setting to App Settings.

Gauravjeetsingh (changes requested):

Gauravjeetsingh:
Also, with multipane workspace we now have another file, which contains this code of intelligent switching. 
It's in `sttm-desktop/www/main/navigator/shabad/utils/change-verse.js`. If you can update it there too, that'd be helpful.

We might get rid of shabadContent.jsx file, in later releases, once we move on to this newer code for all workspaces.

www/configs/user-settings.json
singh-sp: Sure

singh-sp:
@Gauravjeetsingh bhaji, unfortunately, I won't have time (a few weeks) to make those changes. Please feel free to update my branch as desired! 🙏

Gauravjeetsingh:
Closing this out, as we have another PR for this #1989

Issue #1963, opened 2024-06-06
Single display minimize window isn't working


Gauravjeetsingh:
Unable to minimize the search/shabad window in single display mode

Issue #1968, opened 2024-06-06
Preview of viewer pane doesn't change when we change option in dropdown menu from settings


Gauravjeetsingh:
The viewer preview window doesn't change when toggling verse visibility or content dropdown..

Issue #1967, opened 2024-06-06
In quick tools and settings content dropdown, show only available options


Gauravjeetsingh:
Currently it shows option to select punjabi translation for verses, that doesn't have any punjabi translation.

indersinghkhalis:
What's the status here?

Gauravjeetsingh:
Fixed and merged to dev

Issue #1962, opened 2024-06-06
Deleting a custom theme background image throws an error


Gauravjeetsingh:
Unable to delete a custom background image, get an error that internet connection is required even when connected to internet.

Gauravjeetsingh:
Update: this happens when we upload an image that has a space in filename in macos

PR #1969, opened 2024-06-10
Bug fixes and improvements

Gauravjeetsingh:
Fixes issue #1963, #1967, #1968, #1962

### DIFF
--- a/www/main/settings/components/SettingViewer.jsx
+++ b/www/main/settings/components/SettingViewer.jsx
@@ -1,39 +1,22 @@
-import React, { useEffect, useState } from 'react';
+import React from 'react';
 import { useStoreState } from 'easy-peasy';
 
 const themes = require('../../../configs/themes.json');
 
 const SettingViewer = () => {
   const { themeBg } = useStoreState((state) => state.userSettings);
-  const { slideOrder } = useStoreState((state) => state.viewerSettings);
-
-  const [translationOrder, setTranslationOrder] = useState();
-  const [teekaOrder, setTeekaOrder] = useState();
-  const [transliterationOrder, setTransliterationOrder] = useState();
-
-  const orderFunctions = {
-    translation: (item) => {
-      if (translationOrder !== item) {
-        setTranslationOrder(item);
-      }
-    },
-    transliteration: (item) => {
-      if (transliterationOrder !== item) {
-        setTransliterationOrder(item);
-      }
-    },
-    teeka: (item) => {
-      if (teekaOrder !== item) {
-        setTeekaOrder(item);
-      }
-    },
-  };
 
   const {
     gurbaniFontSize,
     content1FontSize,
     content2FontSize,
     content3FontSize,
+    content1Visibility,
+    content2Visibility,
+    content3Visibility,
+    content1,
+    content2,
+    content3,
     displayNextLine,
     leftAlign,
     theme,
@@ -62,15 +45,19 @@ const SettingViewer = () => {
     fontSize: `${gurbaniFontSize * 3}px`,
   };
 
-  const translationStyles = {
+  const content1Styles = {
+    display: content1Visibility ? 'block' : 'none',
     fontSize: `${content1FontSize * 3}px`,
+    fontWeight: 'normal',
   };
 
-  const transliterationStyles = {
+  const content2Styles = {
+    display: content2Visibility ? 'block' : 'none',
     fontSize: `${content2FontSize * 3}px`,
   };
 
-  const teekaStyles = {
+  const content3Styles = {
+    display: content3Visibility ? 'block' : 'none',
     fontSize: `${content3FontSize * 3}px`,
   };
 
@@ -79,12 +66,6 @@ const SettingViewer = () => {
     opacity: 0.5,
     display: displayNextLine ? 'block' : 'none',
   };
-
-  useEffect(() => {
-    slideOrder.forEach((element, index) => {
-      orderFunctions[element](index + 2);
-    });
-  }, [slideOrder]);
 
   const getLarivaarAssistClass = () => {
     if (larivaarAssist) {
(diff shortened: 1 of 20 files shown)


www/main/settings/components/SettingViewer.jsx
```diff
@@ -1,39 +1,22 @@
...
     gurbaniFontSize,
     content1FontSize,
     content2FontSize,
     content3FontSize,
+    content1Visibility,
+    content2Visibility,
+    content3Visibility,
+    content1,
+    content2,
+    content3,
```
theamanjs, line 19: Veerji, what do these content variables represent? 

PS I think this should be renamed to a self-explanatory name

www/main/settings/components/SettingViewer.jsx
```diff
@@ -62,15 +45,19 @@ const SettingViewer = () => {
...
-  const transliterationStyles = {
+  const content2Styles = {
+    display: content2Visibility ? 'block' : 'none',
     fontSize: `${content2FontSize * 3}px`,
   };
 
-  const teekaStyles = {
+  const content3Styles = {
+    display: content3Visibility ? 'block' : 'none',
     fontSize: `${content3FontSize * 3}px`,
```
theamanjs, line 61: Veerji, this piece of code is repetitive. In case of change is required in the future there will be multiple places to change, and to follow the DRY principle, we can refactor this to something like:

```js
const contents = [
  { visibility: content1Visibility, fontSize: content1FontSize },
  { visibility: content2Visibility, fontSize: content2FontSize },
  { visibility: content3Visibility, fontSize: content3FontSize }
];

const contentStyles = contents.map((content, index) => ({
  display: content.visibility ? 'block' : 'none',
  fontSize: `${content.fontSize * 3}px`,
}));

```

www/main/settings/utils/theme-bg-reader.js
```diff
@@ -29,8 +29,8 @@ export const upsertCustomBackgrounds = (responseCallback = () => {}) => {
           .map((file) => ({
             name: file,
             path: `${userBackgroundsPath}/${file}`,
-            'background-image': `${userBackgroundsURL}/${file.replace(/(\s)/g, '\\ ')}`,
-            'background-image-path': `${userBackgroundsPath}/${file.replace(/(\s)/g, '\\ ')}`,
+            'background-image': `${userBackgroundsURL}/${file}`,
+            'background-image-path': `${userBackgroundsPath}/${file}`,
```
theamanjs, line 33: Veerji, this regex was responsible for escaping white spaces in the images that are being uploaded. How are we handling this condition alternatively with this code?

theamanjs (changes requested):

www/main/settings/components/SettingViewer.jsx
Gauravjeetsingh: Initially we had translation, transliteration and teeka. Now instead of that, we are using content1, content2, and content3, because that can be switched.

Context: https://github.com/KhalisFoundation/sttm-desktop/pull/1903

www/main/settings/utils/theme-bg-reader.js
Gauravjeetsingh: fs module handles this on its own. This regex fails in mac, and throws an error. I am still looking for edge cases and would update if needed.

www/main/settings/components/SettingViewer.jsx
Gauravjeetsingh: Looks good, although a little worried about code readability? Because we use these styles at different places, and it would have to be applied like contentStyles[0], contentStyles[1]. 🤔

Gauravjeetsingh:
@theamanjs veeray, I am done with all the changes, can you please review it again.

www/main/settings/utils/theme-bg-reader.js
Gauravjeetsingh: Fixed the edge case with recent commits.

Gauravjeetsingh:
Merging this one for alpha release, @theamanjs veerji, when here, can you review this PR instead
https://github.com/KhalisFoundation/sttm-desktop/pull/1970

www/main/settings/components/SettingViewer.jsx
theamanjs: Veerji I still doubt the naming convention. It would always need an explanation to it's name. We have to come up with better specific name. 

Maybe something like `contentLine1`, `contentLine2`, or `contentLine3` can be a better name. This is just a rough idea to target what exact content we are referring to specifically.

www/main/settings/components/SettingViewer.jsx
theamanjs: Okay, if you are worried about the index count we can refactor it using reduce method.

```js
const contents = [
  { visibility: content1Visibility, fontSize: content1FontSize },
  { visibility: content2Visibility, fontSize: content2FontSize },
  { visibility: content3Visibility, fontSize: content3FontSize }
];

const contentStyles = contents.reduce((acc, content, index) => {
  acc[`content${index + 1}`] = {
    display: content.visibility ? 'block' : 'none',
    fontSize: `${content.fontSize * 3}px`,
  };
  return acc;
}, {});

console.log(contentStyles.content1) // Accessible like this way as well
```

If you have a better approach in your mind to follow the DRY principle we can talk about that as well veerji.

www/main/settings/components/SettingViewer.jsx
Gauravjeetsingh: Moving this to Code refactors issue #1802

www/main/settings/components/SettingViewer.jsx
Gauravjeetsingh: How about this, used a function that returns the style. Initially I thought of contentAttributes as an array instead of object, but then didn't want to abstract it with adding one to index value to get id. In future, if we have another content Id, we can continue with this same logic.

```javascript
  const contentAttributes = {
    1: { visibility: content1Visibility, fontSize: content1FontSize },
    2: { visibility: content2Visibility, fontSize: content2FontSize },
    3: { visibility: content3Visibility, fontSize: content3FontSize },
  };

  const getContentStyles = (contentId) => ({
    display: contentAttributes[contentId].visibility ? 'block' : 'none',
    fontSize: `${contentAttributes[contentId].fontSize * 3}px`,
    fontWeight: 'normal',
  });
```

Issue #1481, opened 2022-03-21
Limit font size using quick tools


theamanjs:
There is option to increase the size of text to any indefinite value using the Quicktools,  resulting in the bani flowing out of the screen. I thing this has been done for 4k display out. - don't know if this is a feature. But the value is capped to 20 in the Settings menu. In My view, the option to increase Bani Size should be capped to 30.

From: slack

PR #2002, opened 2024-07-19
Limit font size for quick tools

yadvirkaur:
#1481
Max font size using quick tools is capped to 30 and min font size to 1.

### DIFF
--- a/www/main/viewer/Slide/QuickTools.jsx
+++ b/www/main/viewer/Slide/QuickTools.jsx
@@ -84,6 +84,8 @@ const QuickTools = ({ isMiscSlide }) => {
     let payload;
     let actionName;
     let stateName;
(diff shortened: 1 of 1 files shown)


www/main/viewer/Slide/QuickTools.jsx
```diff
@@ -84,6 +84,8 @@ const QuickTools = ({ isMiscSlide }) => {
     let payload;
     let actionName;
     let stateName;
+    const maxFontSize = 30;
```
Gauravjeetsingh, line 87: @yadvirkaur let's keep this same as the max font size in settings, change it to 20.

Gauravjeetsingh (changes requested):

Gauravjeetsingh (approved):

Issue #2013, opened 2024-11-12
Allow margin adjustment for screen


saintsoldierx:
Ideally there should also be an option to make the margins, padding to near Zero. To accommodate more text in bigger fonts.

Gauravjeetsingh:
Merged to beta release, closing this out.

PR #2018, opened 2024-12-21
add padding option

gurjit03:
In this PR, we are adding padding tools to adjust the padding in the shabad container 
image


saintsoldierx:
@Gauravjeetsingh @inderpreetsingh can we get this reviewed and merged please?

www/main/common/store/GlobalState.js
```diff
@@ -86,6 +93,33 @@ const GlobalState = createStore({
...
+            actionName: 'setPadding',
+            settingType: 'viewerSettings',
+          }),
+        );
+      }
+      const newState = state;
+      newState.containerPadding[payload.type] = payload.value;
+
+      return newState;
+    }),
```
indersinghkhalis, line 122: can we do something along the lines of

```
const sendUpdateViewerSetting = (target) => {
  target.send(
    'update-viewer-setting',
    JSON.stringify({
      payload,
      actionName: 'setPadding',
      settingType: 'viewerSettings',
    }),
  );
};

if (global.webview) {
  sendUpdateViewerSetting(global.webview);
}

if (global.platform?.ipc) {
  sendUpdateViewerSetting(global.platform.ipc);
}

```

to remove any duplicate code?

indersinghkhalis:
@Gauravjeetsingh lets fasttrack this and get this merged. I have done a high level review from my side, but would love for you to take a look.

Gauravjeetsingh (changes requested):
**Design Considerations for Current Implementation:**
Minor thing: Setting the top margin to 0 causes the "Padding Tools" label to overlap with the padding options menu, which affects readability of the option:
image
Additionally, the label "Padding" may not be intuitive for non-technical users. It might be worth reconsidering the terminology to make its function more self-explanatory.

Referring back to the problem statement outlined in #2013, the primary goal is to reduce padding to accommodate more text on the screen. However, the visual difference between the default padding and setting padding to zero appears minimal:

With default padding: image

With padding set to 0: image

Could we investigate specific shabads or screen sizes where reducing padding meaningfully increases the visible text area? Testing those edge cases would help determine whether this feature effectively addresses the problem.

This investigation would also clarify whether there’s ever a need to increase the right and bottom padding, which are currently defaulted to 0.

gurjit03:
> **Design Considerations for Current Implementation:** Minor thing: Setting the top margin to 0 causes the "Padding Tools" label to overlap with the padding options menu, which affects readability of the option: image Additionally, the label "Padding" may not be intuitive for non-technical users. It might be worth reconsidering the terminology to make its function more self-explanatory.
> 
Yes, that's right. Would it be alright to display the two fields side by side? Also, regarding the naming, I think we can go with spacing or border-spacing. Let me know if you have any other suggestions!

> Referring back to the problem statement outlined in #2013, the primary goal is to reduce padding to accommodate more text on the screen. However, the visual difference between the default padding and setting padding to zero appears minimal:
> 
> With default padding: image
> 
> With padding set to 0: image
> 
> Could we investigate specific shabads or screen sizes where reducing padding meaningfully increases the visible text area? Testing those edge cases would help determine whether this feature effectively addresses the problem.
> This investigation would also clarify whether there’s ever a need to increase the right and bottom padding, which are currently defaulted to 0.
I think that’s a really good observation. From my conversation with @saintsoldierx veerji, the issue seems to be:

“Some Gurdwara projectors are placed quite far from the screen, causing the projected image to go off the edges. So we need a way to add extra padding or margins on any side to fix that.”

To me, it sounds like adding customizable side padding should help manage this. I’ve also added an upper limit to the padding so things don’t get too out of hand.

If you have any other thoughts or suggestions, would love to hear your input!

www/main/common/store/GlobalState.js
gurjit03: @indersinghkhalis thanks ji, will have a look asap.

indersinghkhalis:
I agree with the requirement of customizable padding, while we might do a better job in providing that solution, i thinking providing A solution is better.

indersinghkhalis:
@Gauravjeetsingh  i think padding a standard term for what we are accomplishing here. But yeah let's do all the other changes.

Gauravjeetsingh:
@gurjit03 veerji, Any update on this one?

Gauravjeetsingh:
I am getting blank white screen in viewer after merging changes, and there are few linter issues as well.

Gauravjeetsingh (approved):

PR #2057, opened 2025-06-20
Fix the Aptabase Integration & add sentry for runtime crash logs

Gauravjeetsingh:
Uses a config file for aptabase key as the environment variables are not available in the packaged build.

Also, includes the sentry integration to catch those runtime errors.

### DIFF
--- a/app.js
+++ b/app.js
@@ -25,6 +25,7 @@ const http = require('http-shutdown')(httpBase);
 const io = require('socket.io')(http);
 /* eslint-enable */
 
+const prodConfig = require('./config.prod.json');
 const defaultPrefs = require('./www/configs/defaults.json');
 const themes = require('./www/configs/themes.json');
 const Analytics = require('./analytics');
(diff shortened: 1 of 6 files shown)


packaging/update-packagejson.js
```diff
@@ -10,13 +11,29 @@ if (process.platform !== 'darwin') {
...
+  process.exit(0);
+}
+
 packageJson.build.mac.notarize.teamId = teamId;
+prodConfig.APTABASE_KEY = aptabaseKey;
+prodConfig.SENTRY_DSN = sentryDsn;
 
 fs.writeFileSync('./package.json', JSON.stringify(packageJson, null, 2), 'utf-8');
-console.log('package.json has been updated with the APPLE_TEAM_ID');
+fs.writeFileSync('./config.prod.json', JSON.stringify(prodConfig, null, 2), 'utf-8');
```
gsaran, line 37: Curious, why do we need go this path. If we are able to read these values from `process.env`, why can't these be read directly in app.js?

app.js
```diff
@@ -86,7 +88,31 @@ let startChangelogOpenTimer;
...
+  }
+}
+
+if (aptabaseKey) {
+  aptabase.initialize(aptabaseKey);
+}
+
+if (sentryDsn) {
+  Sentry.init({
+    dsn: sentryDsn,
```
gsaran, line 113: Is Sentry capturing any errors by default once initialized? Or it's usage to capture errors would be a separate PR?

packaging/update-packagejson.js
Gauravjeetsingh: We are able to read these env variables, when the app is being packaged into exe file (for windows) and dmg file (for mac).
But when we install and run that packaged build on our system, these env variables are no longer available to the app.

app.js
Gauravjeetsingh: Sentry can by default catch the following errors:
- Unhandled JS exceptions – anything that bubbles up to the global handlers – are sent automatically.
- Native crashes (seg-faults, out-of-memory, etc.) 

So, for basic setup this works fine, we can create a separate PR (if needed) for manual reporting.

gsaran: Got it. Thanks for the sharing the details.
For the `Unhandled JS exceptions`, I hope we aren't logging something in Sentry that shouldn't be logged like some confidential data.

gsaran (approved):

PR #2096, opened 2025-09-10
Resolve white screen error when connecting external display

Gauravjeetsingh:
Currently, if we connect external display after starting the app, it doesn't work. This fixes that.


indersinghkhalis (approved):
This looks good to me.

PR #2128, opened 2025-12-08
Minor fixes

Gauravjeetsingh:
- Fix the issue of viewer crash when changing themes between video and normal.
- Restore the functionality of deleting specific history entries.

Issue pointed out here:
https://khalis.slack.com/archives/CFS6E54PN/p1765061830910699

Problem:
- For crash, it was creating a root everytime the page loaded. Added an if condition to only create root, when it's null, else directly run the root.render() function.
- For history, needed to add a button that deletes individual items from the history.


amitojsingh (approved):
LGTM

Issue #2193, opened 2026-06-26
Quick Panel crashes when increasing font size for Waheguroo or Mool Mantar slides


uttamdeepk22:
# Problem Statement

The Quick panel has issues with font scaling. Increasing the font size works initially, but after reaching a certain size, the app becomes blank and crashes. This issue is reproducible on my device.

A user also reported that, in the latest version, the Quick Insert ("Waheguru") text can no longer be increased to fill the entire screen. Instead, the text appears to stop scaling at approximately half of the available screen height.

# Steps to Reproduce

1. Select a slide from the Quick Insert panel.
2. Open Quick Panel.
3. Repeatedly increase the font size using the **+** button.
4. Observe the app as the font size increases.

# Expected Behavior

* The Quick Insert text should continue scaling within supported limits without affecting application stability.
* If a maximum font size is reached, the application should gracefully prevent further scaling rather than crashing.

# Actual Behavior

* The Quick panel becomes blank.
* The font size increases normally at first (from that blank quick insert).
* After reaching a certain size. 
* The app becomes blank and crashes.

# Notes

* Reproducible on my device.
* Potentially related to font scaling or rendering limits in the Quick Insert panel. A user has also reported that font scaling no longer fills the screen as it did in previous versions.

# Screenshot:

<img width="725" height="478" alt="Image" src="https://github.com/user-attachments/assets/7d347820-862f-4941-b28a-52cbec1f23ef" />

Issue #196, opened 2018-07-10
Feture-- highlighting shabad line during full shabad display


SukhmanpreetSingh:
![capture](https://user-images.githubusercontent.com/33250386/42488135-c37ff0a0-8421-11e8-8d71-0269247de106.PNG)

AkalUstat:
@maneetpaul are we still doing this?

PR #2199, opened 2026-08-04
Akhand Paatth: continuous scroll

### LINKED ISSUE #1153: Akhand path view auto scroll does not work
In akhand path view if you change the verse from shabad pane and then also scroll from the viewer window, changing to next line on shabad pane does not automatically scroll the viewer window to desired verse.

### LINKED ISSUE #2180: Feature Request: Improve Akhand Paath View by Enabling Continuous/Infinite Scroll
Problem
- A user provided feedback that the current documentation (https://support.khalisfoundation.org/a/solutions/articles/63000076454) lacks sufficient detail and is difficult to understand.
- In the current Akhand Paath view, users cannot continuously scroll through the entire Guru Granth Sahib Ji.
- The view stops at the end of the Ang, requiring users to manually click Next arrow to continue reading.

Feature Request:
- Users should be able to scroll continuously (infinite scroll) through the entire Guru Granth Sahib Ji without interruption using Akhand Path view option.

Additional Context
- The user said that the Infinite/continuous scrolling appears to have existed in earlier versions but is no longer available and added that the current behavior disrupts the reading flow and is not ideal for Akhand Paath use.

amritpal-singh-98:
Closes #2180
Closes #1153

Written with GitHub Copilot. Every commit carries a `Co-authored-by` trailer.

---

## What it does

Akhand Paatth view currently advances one verse at a time, like the slide view.
This changes it to a continuous vertical scroll: the Gurbani flows past a fixed
centre line, and when a Shabad ends the next one is already there. Nobody has to
touch the machine for the reading to continue.

- Speed is set in **verses per second**, not pixels, so the operator's preview
  and the projector stay on the same line despite very different typography.
- The DOM holds a sliding window of a few Shabads. Verses load ahead of the
  reader and are freed behind, so a reading can run for days without growing.
- Play/pause and speed sit in the existing autoplay control. Spacebar toggles.
- Font sizes, translations, transliteration, teeka and spacing all work exactly
  as they do in the slide view, with their own values for this view.

#2180 asked for exactly this and has had no reply since April. #1153 has been
open since 2020: autoplay did nothing at all in this view. It now works — and
"One thing found and deliberately left alone" at the end of this description
shows the line of upstream code that made the automatic advance impossible.

---

## The commits

Eleven, ordered so that each one stands on its own: the tree lints clean and the
suite passes at every commit, and no commit leaves a broken import.

Commits 1–6 are independent of the feature. They are pre-existing fixes
found while building it, and they can be dropped, cherry-picked or taken as a
separate PR without touching the rest.

| # | Commit | Files | What to look for |
|---|---|---|---|
| 1 | `fix`: stop `banidb/constants.js` failing prettier | 1 | `dev` does not lint clean today; this is why. One `prettier-ignore`. |
| 2 | `chore`: add a test setup and run it in CI | 9 | Jest config, first four tests, workflow, lockfile. |
| 3 | `fix`: two long-standing defects in the search pane | 1 | A crash guard, and an IPC listener that was never removed. |
| 4 | `fix`: give native controls back the keys they own | 3 | A focused switch ignored the space bar. |
| 5 | `fix`: keep the Quick Insert tray alive across re-renders | 6 | The "needs two clicks" report. Cause is in `useSlides`. |
| 6 | `fix`: contain a render fault instead of blanking the window | 4 | New error boundary. A render error used to blank the window. |
| 7 | `feat(akhandpatt)`: the scroll engine | 17 | All new, and readable in isolation. Start with its README. |
| 8 | `feat(akhandpatt)`: read across Shabad boundaries | 54 | Where it meets the existing deck. **The riskiest commit.** |
| 9 | `feat(akhandpatt)`: controls, spacing, typography | 27 | Settings and the autoplay control. Mostly threading. |
| 10 | `feat(akhandpatt)`: keys during a continuous reading | 2 | Arrow keys step verses in the slide view only; Space toggles. |
| 11 | `docs(akhandpatt)`: revise code comments and documentation | 38 | Comments and test names only. No code changes; see below. |

Commit 11 changes no executable code. Stripping every comment with Babel leaves
the shipped source byte-identical to commit 10, and the only differences in the
test files are `it()` titles. Both checks carry a control that fails on an
injected change, so a clean result means the check ran. Reviewing commits 1–10
covers the whole of the code.

---

## How to review this

It is a large diff and it does not need reading end to end. Suggested order,
about 40 minutes for the parts that matter:

| Read | Why |
|---|---|
| `www/main/viewer/akhandpatt/README.md` | The design, and why each alternative was not taken. Start here. |
| `.../akhandpatt/scroll-config.js` | Every tunable constant with the measurement behind its value. |
| `.../akhandpatt/shabad-window.js` | The sliding window: pure functions, no DOM. |
| `.../akhandpatt/useAkhandpattScroll.js` | The driver. Its docstring carries a map of the file. |
| `www/main/viewer/ShabadDeck/ShabadDeck.jsx` | Where it meets the existing deck. |

Everything else is a control, a setting, or a test.

Size, honestly: 124 files. 39 of those are tests (+6,801 lines) and 11 are
the new module (+2,644). That leaves **73 existing files carrying 2,671 changed
lines between them**, and most are small: 12 change by five lines or fewer, 38 by
thirty or fewer. The largest change to an existing file is `ShabadDeck.jsx` at
283 lines. The one genuinely large new file is the scroll driver, 1,812 lines,
which carries a map of itself at the top.

On comment density: the new module is 41% comment against 7% across the rest of
`www/main`. That is a deliberate choice for one folder and not a house style
proposal — the constants note the measurement behind them, and the frame loop
explains reasoning you cannot see in the code. Happy to thin it if you would
rather it matched the surrounding files.

---

## Tests

```
npm test          # eslint + htmlhint + jest
npx jest          # just the unit tests
```

388 tests across 35 suites. There were none before this PR. CI runs them on
every pull request and on pushes to `dev`, `master` and `release`
(`.github/workflows/test.yml`).

The three build workflows are red before this PR and stay red: they fail
identically on the unmodified base commit. `berry-build.yml` is a malformed
workflow file, the macOS build needs code-signing secrets, and the Windows build
breaks because the runner image now ships Visual Studio 18, which the pinned
node-gyp does not recognise. None of them are touched here — the only workflow
file this PR changes is the one it adds.

Two notes on how the tests are written.

Assertions were checked by breaking the code: where a test guards a specific
bug, the fix was reverted to confirm the test goes red. Noted in the commit
message wherever it was done.

Thirteen suites read source or config text instead of executing it. That is
limited to properties you cannot assert any other way — a JavaScript constant
that has to match an SCSS variable, the settings list the components read, the
import boundary, and the rule every route into the viewer follows. Each such
suite first asserts that its scan still finds something, so a regex that quietly
stopped matching fails rather than passing vacuously. Tests that merely asserted
a call was present have been removed; that is a real weakness, and it let
regressions through during this work.

---

## Trying it by hand — five minutes

1. Settings → check **Akhand Paatth View**.
2. Search any Shabad and open it. It should start reading down the page.
3. Let it run to the end of the Shabad. The next one should arrive without a
   click. (Turn the speed up to reach the boundary faster.)
4. Press **space**. It stops. Press it again. It resumes from where it stopped.
5. Drag the speed slider, and use the **+**/**−** steppers either side.
6. Scroll the wheel over the deck. It takes over, then hands back to the
   autoscroll after a beat.
7. Open a second display (Presentation mode). Both windows should show the same
   line, including after a wheel nudge in the preview.
8. Change the Bani font size and the two spacing sliders while it reads. The line
   you were on should stay on the centre line.
9. Turn Akhand Paatth off. The slide view should behave exactly as before.

---

## Changes that will look unrelated

Six store files stop returning the immer draft. `GlobalState.js` (7 places),
`ViewerState.js` (3), `OverlayState.js`, `create-navigator-settings.js`,
`create-user-settings-state.js` and `create-overlay-settings-state.js` each ended
an action with `return state`. easy-peasy's `simpleProduce` assigns whatever an
action returns onto the previous committed state and then revokes the draft, so
that line leaves a revoked proxy in the store and every later dispatch throws
`Cannot perform 'get' on a proxy that has been revoked`. It is latent at `dev`
because those particular actions are rarely dispatched twice in a session; the
continuous scroll dispatches them constantly. `test/store/no-draft-return.test.js`
pins it, and that suite fails on a clean `dev` checkout, so it is fixing a real
upstream bug. (A plain `useReducer` in
`navigator/misc/state-manager/reducer.js` still returns `state`, correctly; the
suite only covers easy-peasy actions.)

A settings label was renamed from `AKHAND_PATH_VIEW` to
`AKHAND_PAATTH_VIEW` (and the same for `..._UNAVAILABLE`), matching the spelling
already used in the label text. These are i18n lookup keys for the settings
pane, not the stored setting name, which is unchanged, so no user's saved
settings are affected. `www/configs/user-settings.json` and `www/locales/en.json`
are the only two files involved, and `en.json` is the only locale file in the repo.

Three smaller ones:

- `www/main/banidb/constants.js` adds `prettier-ignore` to the Gurmukhi
  transliteration tables. Those tables violate prettier today; the rule reports
  three errors on a clean `dev` checkout. `prettier-ignore` fixes them without
  reflowing the columns, which are aligned deliberately.
- `www/main/common/sttm-ui/inputbox/InputBox.jsx` loses `handleSpace`. Spacebar
  handling moved into `useKeys`, which now decides which control owns a key.
  Nothing is lost: typing a space in the search box still types a space, and a
  test covers it.
- `package-lock.json` will be reported by GitHub at roughly ±11,000 lines. The
  real change is **+2,790/−3**: 197 new packages, none removed. They are `jest`,
  `jest-environment-jsdom` and `babel-jest` with their transitive dependencies,
  all `devDependencies`. **No runtime dependency is added or changed**, so the
  shipped bundle is unaffected. The line-count inflation is a diff-alignment
  artefact of new entries interleaving alphabetically among 2,220 identical ones;
  `git diff --histogram` shows the true figure.

---

## What this does not do

- Highlighting the current line (#196) is untouched.
- The Quick Panel font-size crash (#2193) is untouched.
- A projection window opened *after* a reading has started begins at the
  operator's last selection rather than the current line.
- The wheel glide is tuned per frame, so it travels further per gesture on a
  high-refresh display.
- Search results and favourites do not re-centre a verse that is already open if
  the reading has scrolled away from it. Re-opening from the Shabad pane does.


amritpal-singh-98:
@Gauravjeetsingh, please could you advise on who could review and approve?

